PR #33, opened 2014-06-05
CSPify stylesheets in the same way as for javascript files

pdelanauze:
For use in chrome extensions, i needed to un-inline my styles into a stylesheet, just like for javascript files .. Came up with this proof of concept which works with Canary and latest versions of Polymer .. 

Opening a pull request for feedback, not sure if this is something that should be merged in the core , if so , maybe some better, more verbose configuration options would be necessary, since the csp: true option will have a different behaviour with style tags under this branch..

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -174,7 +174,7 @@ function handleMainDocument() {
     inlineScripts($, options.outputDir);
   }
 
-  // strip scripts into a separate file
+  // strip scripts and stylesheets into a separate file
   if (options.csp) {
     if (options.verbose) {
       console.log('Separating scripts into separate file');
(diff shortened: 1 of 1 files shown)


Shadowfaxenator:
Any updates on this issue?

pdelanauze:
Haven't had the chance to work on this again , and its now very outdated .. Think we should close this and someone else give it another shot ? if i do land on a project that requires this i'll gladly take another stab at this

azakus:
That sounds good @pdelanauze, I really appreciate the effort on this, even when I haven't had the time to focus on it specifically. FWIW, Polymer 0.8+ has a completely different method of using styles, and that will require a different strategy.

ankit2711:
Is there a solution for this issue within Polymer or any other build tool? What is the status of support for CSP including styles within Polymer?

PR #38, opened 2014-07-02
Feature/strip excludes

shaohua:
- Add "stripExcludes" option in config.json
- Excluded tags are stripped completely when "stripExcludes" is true.
- Useful for excluding local jQuery in favor of CDN.
- CLA signed but it might take a while to process


azakus:
Vulcanize has a strip excludes now. Closing this PR.

PR #39, opened 2014-07-05
I took a shot at making this run asynch

wballard:
Main thing, I use less and litcoffee rather a lot, so I did some re-arranging to make this asynch to allow filter style plugins. Most other node modules are asynch, so I felt the need to go asynch in order to attach modules like browserify, which I use to build my polymer element scripts.


azakus:
Vulcanize is async now. Closing this PR.

PR #80, opened 2014-10-20
Allow passing in options directly to lib/vulcan.

wprl:
This allows separating the engine from the CLI for tools (such as spock)
that want to write files in a different manner or not at all, e.g. for
streaming into later processors in a gulp build step.

### DIFF
--- a/bin/vulcanize
+++ b/bin/vulcanize
@@ -13,6 +13,7 @@ var path = require('path');
 var fs = require('fs');
 var nopt = require('nopt');
 var vulcan = require('../lib/vulcan.js');
+var optparser = require('../lib/optparser.js');
 
 var help = [
   'vulcanize: Concatenate a set of Web Components into one file',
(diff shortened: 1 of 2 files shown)


azakus:
The concept of this PR is now in Vulcanize. Closing this PR.

wprl:
One thing that is in this PR, but not added to vulcanize (or at least not mentioned in the README) is the ability to process file content that may not reside on a file system at a path. For example with gulp, it is much more flexible without that constraint. Often I do not want to write files to disk between processing steps, as it is inefficient, requires more of the gulp processing pipeline to be synchronous, and is prone to error.

I would offer to create a separate PR to add that feature, but given the pace at which this PR progressed I have to say I am highly discouraged.

azakus:
This is also possible with the `loader` option. The API contract is admittedly not documented, but it is very simple: https://github.com/Polymer/hydrolysis/blob/master/lib/loader/fs-resolver.js

A class with one function `accept` that takes a URL and a deferred promise.

wprl:
Ok cool. I am placated :)

PR #123, opened 2015-01-04
Remove unused script name of CSP config

ragingwind:
CSP options doesn't need to have filename replaced. It's a duplicate job with https://github.com/Polymer/vulcanize/blob/master/lib/vulcan.js#L310


ragingwind:
I was missing something.

PR #144, opened 2015-03-17
Don't strip white space; rather collapse to a single space.

trevordixon:
This is a potential fix for https://github.com/Polymer/vulcanize/issues/63 which has been an irritant to me and my team for some time.

Rather than removing empty text nodes, consecutive white space in text nodes is replaced with a single space—more like what is specified in the HTML spec.

I see two main deficiencies with what I'm submitting here:

1. White space in `pre` and `textarea` elements isn't collapsed, but there may be other elements that deserve the same exception, and, in fact, the CSS `white-space` rule can be used to make white space significant for any element. @azakus' suggestion in https://github.com/Polymer/vulcanize/issues/63#issue-41528475 is a solution to this problem.

  The current approach of removing all text nodes that consist only of white space is no better in this regard.

2. If a text node ends with a space and the next node is a text node that begins with a space, those consecutive white space characters won't get collapsed. Instead, consecutive white space characters appear in the vulcanized output. This wouldn't be terribly hard to fix, but I don't think it's a big deal. I opted to keep this pull request simple, but I'll fix it if requested.

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -175,7 +175,7 @@ function findScriptLocation($) {
   return pos;
 }
 
-function isCommentOrEmptyTextNode(node) {
+function isComment(node) {
   var d = node.data;
   if (node.type === 'comment') {
     var keep = (/@license/.test(d));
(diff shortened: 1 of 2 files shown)


azakus:
LGTM. I think this is probably as far as I'd be willing to go for this. Trying to evaluate CSS rules to check for whitespace or inline displays is probably in the realm of a different tool.

Thanks!

PR #149, opened 2015-04-08
Rewrite for Polymer 0.8 elements

azakus:
Complete rewrite on top of [hydrolysis](https://github.com/PolymerLabs/hydrolysis) and [dom5](https://github.com/PolymerLabs/dom5) for Polymer 0.8 style elements and better code health.

CSP functionality is now handled by [crisper](https://github.com/PolymerLabs/crisper)

A Simple gulp plugin will follow in later versions.

Merges for polymer 0.5 elements should be based on the polymer-0.5 branch, which will be maintenance only.


ragingwind:
Great job thanks. I have to use it for 0.8 now

PR #152, opened 2015-04-10
[0.8] Inlining support for <link rel="import" type="css" ...>

nevir:


peterwmwong:
@nevir Do you know if/when this eventually be supported?  I noticed quite a few 0.8 paper elements rely on imported css... and don't look quite right when I vulcanize them :)

azakus:
@peterwmwong Landing support for CSS inlining soon in hydrolysis.

matheusdavidson:
im having the same problem!
Tried to use paper-checkbox but nothing was on the page and paper-ripple covers all page, after some debugging found that paper-checkbox imports paper-checkbox.css and in my vulcanize.html there is no <style> wrapper on the css../

Im using broccoli-vulcanize(for emberjs), is there any way to fix this?

azakus:
@matheusdavidson I think broccoli-vulcanize 2.0 (released yesterday) addresses your problem. At least, it uses vulcanize `^1.4`, where we correctly handle the `type="css"` imports.

mbykovskyy:
@azakus Does the content of the imported css get concatenated with the rest of the HTML? Because my test produces this.

```html
<!DOCTYPE html><html><head>
    <meta charset="utf-8">
    <link rel="import" type="css" href="style.css">
</head>
  <body>
<div hidden="" by-vulcanize=""><dom-module id="existing-component" assetpath="/">
  <template>
    <content></content>
  </template>
</dom-module>
<script>
  Polymer({is: 'existing-component'});
</script>
</div>
    <existing-component>Hello Import!</existing-component>
    <non-existing-component></non-existing-component>
</body>
</html>
```

See how `<link rel="import" type="css" href="style.css">` gets put into the header?

mbykovskyy:
@azakus Never mind, I missed the `inlineCss` option. ;)

PR #154, opened 2015-04-12
Add implicit exclude stripping.

garlicnation:


azakus:
LGTM. Can you update the README as well?

PR #162, opened 2015-04-23
Make all tests pass on Microsoft Windows

MetaMemoryT:
Tests are only failing on nodejs 0.10, because of dirname method, is 0.10 still supported?

### DIFF
--- a/lib/pathresolver.js
+++ b/lib/pathresolver.js
@@ -11,7 +11,7 @@
 // jshint node:true
 'use strict';
 
-var path = require('path');
+var path = require('path').posix;
 var url = require('url');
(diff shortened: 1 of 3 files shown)


azakus:
Looks like this breaks on node 0.10, because the `path.posix` is a new api.
@garlicnation, thoughts on an alternative?

MetaMemoryT:
Oh I just learned about an alternative from my other PR, see comments on https://github.com/sindresorhus/gulp-vulcanize/pull/24

azakus:
Merge of #199 made this PR not work anymore. Implemented with 702cfb1

PR #166, opened 2015-04-24
finalize comment stripping.

garlicnation:


lib/vulcan.js
```diff
@@ -126,6 +126,18 @@ function flatten(tree, bodyFragment) {
       }
     }
   }
+  // Deduplicate comments
+  var commentsByData = {};
+  dom5.nodeWalkAll(doc, dom5.isCommentNode).forEach(function(comment) {
+    commentsByData[comment.data] = comment;
```
azakus: In 0.7.x, i removed all whitespace to account for the difference between

    

and

garlicnation: Resolved.

azakus:
LGTM after squashing

PR #177, opened 2015-05-13
Readme looks great - just added a section at the top...

tjsavage:
...about why you might want to use Vulcanize.


azakus:
I'm going to add a section about HTTP/2 and Server Push obsoleting vulcanize in the fullness of time.

PR #178, opened 2015-05-14
Fix inline script excludes

### LINKED ISSUE #158: [1.2.1] Using --inline-scripts and --exclude
I have excluded a path containing a js script, and combined with -inline-scripts.
The html file generated contains:  <script></script>, I think this is wrong because that line does nothing.
I expected vulcanize to do nothing with that line, it just removed the src attribute.

Thanks!

garlicnation:
Fixes #158

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -34,10 +34,13 @@ function reparent(newParent) {
 }
 
 function isExcludedImport(importMeta) {
+  return isExcludedHref(importMeta.href);
+}
+
+function isExcludedHref(href) {
   if (!Options.excludes) {
     return false;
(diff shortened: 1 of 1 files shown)


azakus:
LGTM

lib/vulcan.js
```diff
@@ -182,6 +185,9 @@ function inlineScripts(doc, href) {
     var src = dom5.getAttribute(script, 'src');
     var uri = URL.resolve(href, src);
     // let the loader handle the requests
+    if (isEcludedHref(src)) {
```
azakus: `isExcludedHref`

Issue #195, opened 2015-06-01
Slashes in URLs are converted into backslashes on Windows


achimnol:
I'm using node.js 0.12.4 on Windows 8.1 64bit and have installed vulcanize by `npm install -g vulcanize` as instructed. The compilation process goes fine without warnings nor errors.

Surprisingly, I see some images and CSS styles are missing after vulcanizing because it has converted all slashes in every URL into backslashes! 
For example: `background-image: url(somepath/somefile.png)` => `background-image: url(somepath\somefile.png)` and the browser is trying to load "somepathsomefile.png". In some contexts, they are converted again into percent-encoding, `%5C`. 
Loading the original codes in the browser works perfectly fine but only the vulcanized code has missing resources like this. This issue does not happen on Mac and Linux.

PR #196, opened 2015-06-01
Always use POSIX-style path functions to keep slashes intact.

achimnol:
This is a simple fix for #195. I have confirmed this resolves the issue on Windows.


achimnol:
I signed it!

garlicnation:
This doesn't work in node 0.10 :'(. Maybe add a check to see if path.posix exists and use that if available?

azakus:
Looks like some kind soul made a 0.10 compatible [path-posix](https://www.npmjs.com/package/path-posix) module by just extracting that portion from 0.12+

azakus:
The merge conflict was a bit confusing after merging #199, so I did it manually in https://github.com/Polymer/vulcanize/commit/774e43ca8ab26c4e1bab2a5c98ee96fd3b90fcaf.

Thanks!

PR #199, opened 2015-06-04
Class-based vulcan API

### LINKED ISSUE #190: Vulcanize does not support concurrent calls
Vulcanize has a global options setting that does not play well when used concurrently. This is most easily seen when using the gulp-vulcanize plugin.

We'll probably need a breaking-change refactor for this.

garlicnation:
Fixes #190


azakus:
The deprecated API is a nice way forward!

azakus:
LGTM, this will be 1.9.0

PR #207, opened 2015-06-17
Bump hydrolysis version.

garlicnation:
Bump hydrolysis version.


azakus:
Shouldn't be strictly necessary, but LGTM

PR #221, opened 2015-07-25
fix process function

tikicoder:
The change in the code is proposed because you should be able to pass an empty string and it process as null.  Plus the issue is some site showing how to use the function, for gulp for example, https://www.npmjs.com/package/gulp-vulcanize, show it as an empty string.  

The better way would be to trim the sides of whitespace so you could do  ' ' and it still return null.

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -51,7 +51,7 @@ function buildLoader(abspath, excludes) {
 var Vulcan = function Vulcan(opts) {
     // implicitStrip should be true by default
   this.implicitStrip = opts.implicitStrip === undefined ? true : Boolean(opts.implicitStrip);
-  this.abspath = (String(opts.abspath) === opts.abspath && (opts.abspath!==null && String(opts.abspath).trim()!=='')) ? path.resolve(opts.abspath) : '';
+  this.abspath = (String(opts.abspath) === opts.abspath && String(opts.abspath).trim() !== '') ? path.resolve(opts.abspath) : '';
   this.pathResolver = new PathResolver(this.abspath);
   this.excludes = Array.isArray(opts.excludes) ? opts.excludes : [];
   this.stripExcludes = Array.isArray(opts.stripExcludes) ? opts.stripExcludes : [];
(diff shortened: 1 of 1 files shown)


tikicoder:
I signed it

azakus:
It's probably easier to just check the truthiness of abspath:
```javascript
this.abspath = (opts.abspath && String(opts.abspath) === opts.abspath) //...
```

tikicoder:
That might work now but really is not a good way.  What happens if js became stricter.  An empty string is not true or false.  Plus if you do it that way a string that is just spaces is true.  The way I suggested could be done to trim the white spaces on the ends.  What happens if js gets updated so an empty string is true.  Letting the compiler determine how to process something can cause more issues.

azakus:
That scenario will probably never happen; it would break too many things to change the default truthiness of primitives like empty string. At most, a new pragma like `'use strictest'` could be to be introduced to change behavior this way, but that would probably break even more things than this anyway.

tikicoder:
It would.  But why do something that should be concidered bad programming.  If you were doing this in other languages would you test the truthfulness of a string or would you test the string.  Again, why write code you wouldn't do in other languages is js just because you can.  I would never do that in c#/php/c++.  When I see it in vb.net I turn on option explicit and option script and fix it.

lib/vulcan.js
```diff
@@ -51,7 +51,7 @@ function buildLoader(abspath, excludes) {
 var Vulcan = function Vulcan(opts) {
     // implicitStrip should be true by default
   this.implicitStrip = opts.implicitStrip === undefined ? true : Boolean(opts.implicitStrip);
-  this.abspath = String(opts.abspath) === opts.abspath ? path.resolve(opts.abspath) : '';
+  this.abspath = (String(opts.abspath) === opts.abspath && (opts.abspath!==null && String(opts.abspath).trim()!=='')) ? path.resolve(opts.abspath) : '';
```
azakus: The `opts.abspath !== null` is redundant, as `String(x) === x` if and only if x is a string.

Put spaces between the equality check: `.trim() !== ''`

azakus:
LGTM

I'll merge after nits are fixed.

tikicoder:
I have made the requested changes.  

I agree the opts.abspath !== null is redundant, but that is only because String(null) appears to return an empty string.  Since null is a valid value for string it technically could be a string.  Which is why in most other languages I have dealt with I check string != null, empty, or white space.   Which normally ends up being a function that does something to the affect
string != null && string.trim() != '' 

I know in some languages the null is a character, \0 if I remember right.

azakus:
LGTM

PR #227, opened 2015-08-07
Detect polymer css imports and set the 'is' attribute as custom-style.

birkir:
This fixes web component (Polymer) imports for css types. Adds the is="custom-style" attribute.


stramel:
@usergenic I think this PR is way out of date and type="css" has been deprecated, correct?

birkir:
You're right, this PR is 2 years old

PR #232, opened 2015-08-20
Don't search js files for transitive dependencies.

garlicnation:


azakus:
LGTM, we can add a test later

PR #247, opened 2015-09-23
Fix inlining and absolute paths on windows

### LINKED ISSUE #244: Inline Scripts/CSS Error
In version ```1.10.1``` everything works just fine but when I update to any version after that I begin to see problems. This is my input:
```
vulcanize -p "." Components/dvr-manager/dvr-manager.html > whatever.html --inline-scripts
```

This is the error I get:
```
{ [Error: ENOENT, open 'c:\projects\byutv2\Webs\Cms.Web\projects\byutv2\Webs\Cms.Web\null\bower_components\web-animations-js\web-animations-next-lite.min.js']
   errno: -4058,
   code: 'ENOENT',
   path: 'c:\\projects\\byutv2\\Webs\Cms.Web\\projects\\byutv2\\Webs\\Cms.Webs\\null\\bower_components\\web-animations-js\\web-animations-next-lite.min.js' }
```

It looks like its creating the path to the file incorrectly. It repeats some portions of it and even throws a ```null``` in there. This happens the same for the ```--inline-scripts``` flag and the ```--inline-css``` flag.

azakus:
Fixes #244

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -222,8 +222,7 @@ Vulcan.prototype = {
       if (this.isExcludedHref(src)) {
         return Promise.resolve(true);
       }
-      var filePath = path.resolve(this.pathResolver.urlToPath(uri));
-      return this.loader.request(filePath).then(function(content) {
+      return this.loader.request(uri).then(function(content) {
         if (content) {
           dom5.removeAttribute(script, 'src');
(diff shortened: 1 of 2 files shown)


azakus:
Don't know why I added the `path.resolve` in the inlining, but removing that line fixed it :'(

PR #252, opened 2015-09-28
Path hacks to support bower_components

garlicnation:


azakus:
This seems to break some tests. Do those tests need to be refactored?

garlicnation:
I don't know, I basically broke the pathresolver to make sharding work, but I'm not 100% clear on the original goals of the path resolver.

azakus:
This was resolved with the other PRs right? Closing this one then.

PR #257, opened 2015-10-01
"fix" excludes with the --redirect flag.

garlicnation:
HTML imports are still broken, and will require a more nuanced approach to the
NoopResolver. Since the NoopResolver has no visilbity into redirected URLs,
it will have a hard time matching the custom://protocol/url type links.


azakus:
Lgtm

PR #260, opened 2015-10-05
Strip Excludes should be fuzzy like Excludes

### LINKED ISSUE #254: --strip-exclude Doesn't work for relative paths
To have vulcanize successfully strip a `<link>` tag, the path has to be absolute. 

Suggest using `href.search(r) >= 0` instead of `href == r` inside the `isStrippedImport` function (as it is in the `isExcludedHref` function).

azakus:
Fixes #254 --strip-exclude Doesn't work for relative paths

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
   - "0.10"
+  - "0.12"
   - "4"
 sudo: false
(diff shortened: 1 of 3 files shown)


azakus:
@garlicnation PTAL

garlicnation:
LGTM

PR #281, opened 2015-11-28
Escape inline scripts

### LINKED ISSUE #280: XSS vulnerability with script inlining
Proof of concept: https://gist.github.com/TimothyGu/f5d7c86d248de4ed08a2

The original HTML file defines `b` as a variable, while the transformed `out.html` pops up an alert box with content "XSS".

Outside of a security context, this bug makes it impossible for users to use firebase elements with vulcanize, as reported by @johnwhitton on the Polymer Slack.

azakus:
Use portion of UglifyJS2 to handle escaping inline-scripts
Modified UglifyJS2 located in third_party/UglifyJS2/output.js
Modifications listed in third_party/UglifyJS2/README.md

Added test case from @TimothyGu
Use chai for assert testing

Fixes #280

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -20,6 +20,7 @@ var CommentMap = require('./comment-map');
 var constants = require('./constants');
 var matchers = require('./matchers');
 var PathResolver = require('./pathresolver');
+var encodeString = require('../third_party/UglifyJS2/output');
 
 var Promise = global.Promise || require('es6-promise').Promise;
 
@@ -265,6 +266,7 @@ Vulcan.prototype = {
       }
       return this.loader.request(uri).then(function(content) {
         if (content) {
+          content = encodeString(content);
           dom5.removeAttribute(script, 'src');
           dom5.setTextContent(script, content);
         }
--- a/third_party/UglifyJS2/README.md
+++ b/third_party/UglifyJS2/README.md
@@ -1,0 +1,17 @@
+# Contents
+Slim copy of [UglifyJS2](https://github.com/mishoo/UglifyJS2) output library.
+
+# Original Copy
+https://raw.githubusercontent.com/mishoo/UglifyJS2/v2.6.1/lib/output.js
+
+# Modifications
+
+- Removed all functions except `encode_string`.
+- Replaced call `var ret = make_string(str, quote)` with `var ret = str`
+- Added default `options` object
+- Added CommonJS `module.exports` line
+- Added `jshint` settins line
+- Removed semicolon from defintion of `encode_string`
+
+# License
+BSD-2-Clause: [License](https://raw.githubusercontent.com/mishoo/UglifyJS2/v2.6.1/LICENSE)
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "path-posix": "^1.0.0"
   },
   "devDependencies": {
+    "chai": "^3.4.1",
     "jshint": "^2.7.0",
     "mocha": "^2.2.4"
   },
--- a/test/html/xss.html
+++ b/test/html/xss.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<!--
+    @license
+    Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+    This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+    The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+    The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+    Code distributed by Google as part of the polymer project is also
+    subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<!--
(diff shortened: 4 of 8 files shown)


TimothyGu:
This doesn't fix the problem. The new output is

```html
<script>var b = 0<\/script><script>alert('XSS'); //2;
</script>
```

Which as you can see is not exact valid JavaScript.

The terrible thing about this piece of script is that you can't just JavaScript-escape that slash since it is not in a string, but rather the starting of a regex.

TimothyGu:
On this note, UglifyJS does things correctly:

```js
var b=/script><script>alert('XSS'); / /2>0
// which is equivalent to 0</script><script>alert('XSS'); //2
```

TimothyGu:
Actually scratch that: that reordering is due to UglifyJS compressor's option `comparison` being on. When I turned that off this is returned:

```js
var b=0</script><script>alert('xss'); / /2
```

Which while doesn't have an immediate danger, opens up the possibility for a more sophisticated attack like

```js
var g = 1;
var b = 0</script><script>alert('xss'); //g;
```

After uglifying (without `evaluate` and `comparison` _compressor_ options on), this is the output:

```html
<html><head><meta charset="UTF-8"><script>var g=1,b=0</script><script>alert('xss'); / /g
</script>
</head><body></body></html>
```

I'd understand if you guys don't want to fix this; it is the rarest of the rarest case, but if vulcanize should be truly safe against untrusted input this is an impediment that is pretty difficult to fix (as I have already and you are starting to) find out.

azakus:
@TimothyGu The option you want for uglify is `--inline-script`, which will escape any `</script>` line in a javascript line.

For your example in https://github.com/Polymer/vulcanize/pull/281#issuecomment-160253679, the PR would make this output

```html
<!DOCTYPE html><html><head><meta charset="UTF-8"><script>var g = 1;
var b = 0 <\/script><script>alert('xss'); / /g;
</script>
</head><body></body></html>
```
Which throws a `SyntaxError`. I think this should be sufficient for this case, and it is probably the best that can be done without using another tool that does a more active analysis.

azakus:
@garlicnation PTAL

TimothyGu:
Yes, my last example does show a SyntaxError: but that's exactly the bug. By itself, the JavaScript is completely valid. When embedded into HTML (after vulcanization) it's not.

azakus:
Well yes, the script now has a SyntaxError because it was embedded into HTML. There is no fixing that, it is baked into the HTML parser spec: https://html.spec.whatwg.org/multipage/scripting.html#restrictions-for-contents-of-script-elements.

The `--inline-scripts` flag of vulcanize is off by default, and will remain that way. If you want to bundle the script portions of elements your own way, you can run `vulcanize` without the `inline-scripts` flag and pull out the script portions into your own bundle as you see fit.

If you use `dom5`, what vulcanize uses, you can do it with a simple tree walker

Example:
```js
var doc = dom5.parse(documentContent);
var scriptContents = [];
dom5.queryAll(doc, dom5.predicates.hasTagName('script')).forEach(function(script) {
  var content = dom5.getTextContent(script);
  if (dom5.hasAttribute(script, 'src')) {
    content = fs.readFileSync(dom5.getAttribute(script, 'src'), 'utf-8');
  }
  scriptContents.push(content);
  dom5.remove(script);
});
var scriptElement = dom5.constructors.element('script');
fs.writeFileSync('script.js', scriptContents.join('\n'), 'utf-8');
dom5.setAttribute(scriptElement, 'src', 'script.js');
dom5.append(dom5.query(doc, dom5.predicates.hasTagName('body')), scriptElement);
documentContent = dom5.serialize(doc);
```

TimothyGu:
Basically you are saying that there is no way of fixing the more complex version of XSS without extensive processing. Very well. This PR at least makes `</script>` in a string work, so better than nothing I guess.

third_party/UglifyJS2/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+# Original Copy
+https://raw.githubusercontent.com/mishoo/UglifyJS2/v2.6.1/lib/output.js
+
+# Modifications
+
+- Removed all functions except `encode_string`.
+- Replaced call `var ret = make_string(str, quote)` with `var ret = str`
+- Added default `options` object
+- Added CommonJS `module.exports` line
```
TimothyGu: `module.exports` itself is an invention of Node.js. CommonJS Modules/1.1 spec says that only the plain `exports` variable is allowed to be the means of exporting things.

lib/vulcan.js
```diff
@@ -265,6 +266,7 @@ Vulcan.prototype = {
       }
       return this.loader.request(uri).then(function(content) {
         if (content) {
+          content = encodeString(content);
```
garlicnation: Should this be flag controlled?

garlicnation:
LGTM

lib/vulcan.js
azakus: I don't think so. If a user is inlining scripts, I think they are going to expect this behavior.

PR #282, opened 2015-11-30
excludes does not accept regexes

### LINKED ISSUE #266: `excludes` doesn't accept RegExp[] as described in the readme
I am new to Polymer and I'm trying to use [broccoli-vulcanize](https://github.com/mbykovskyy/broccoli-vulcanize) in my Ember.js app. I was really confused when I tried to compile with the example setting and it failed.

After spending some time digging around the source code of broccoli-vulcanize and vulcanize, it appears to me that vulcanize had changed the argument type for `excludes` from an Array of `RegExp` objects to an Array of String, without bumping the major version to reflect a breaking change.

I think the [readme](https://github.com/Polymer/vulcanize#using-vulcanize-programmatically) should really be updated to reflect the changes.

franzliedke:
An update to the README that would fix #266.

Apparently, due to the use of `.match` on these array items (in vulcan.js), they are expected to be strings and not `RegExp` objects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Vulcanize as a library has two exported function.
 options.
 - `abspath`: A folder to treat as "webroot".
   - When specified, use an absolute path to `target`.
-- `excludes`: An array of RegExp objects to exclude paths from being inlined.
+- `excludes`: An array of strings with regular expressions to exclude paths from being inlined.
 - `stripExcludes`: Similar to `excludes`, but strips the imports from the output entirely.
     - If `stripExcludes` is empty, it will be set the value of `excludes` by default.
 - `inlineScripts`: Inline external scripts.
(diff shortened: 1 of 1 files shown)


franzliedke:
I signed it!

azakus:
LGTM
PTAL @garlicnation

franzliedke:
Not one hundred percent sure whether my examples are correct. My according [pull request](https://github.com/mbykovskyy/broccoli-vulcanize/pull/8) to broccoli-vulcanize is failing tests. :-/

LeondidV:
Hello azaks! SoS! SoS!
I am a beginner of Polymer programmer. I have a problem programming Polymer.
I had a Polymer program. But during building, the error is checked.
Error is followin as:
 "This version of vulcanize is not compatible with polymer<0.0. Please use vulcanize version <0.7.x.
File:\\......"
I don't understand that. I did following your suggestion.
But this error is still checked. Help me!
Best regards.
Leonid.

PR #285, opened 2015-12-16
Include example for exclude syntax

robdodson:


README.md
```diff
@@ -25,7 +25,7 @@ for this step).
 - `-h`|`--help`: print this message
 - `-v`|`--version`: print version number
 - `-p <arg>`|`--abspath <arg>`: use <arg> as the "webserver root", make all adjusted urls absolute
-- `--exclude <path>`: exclude a subpath from root. Use multiple times to exclude multiple paths. Tags to excluded paths are kept.
+- `--exclude <path>`: exclude a subpath from root. Use multiple times to exclude multiple paths. Tags to excluded paths are kept. ex: `--exclude=elements/x-foo.html`
```
ebidel: How about a multiple exampel: `--exclude=elements/x-foo.html --exclude=elements/x-foo2.html`

ebidel: Can we update to a better description. 

"Tags (imports/scripts/etc) that reference an excluded path are left in-place, meaning the resources are not inlined"

robdodson:
Updated. @ebidel PTAL

azakus:
LGTM

PR #293, opened 2016-01-11
Fix ordering of import dependencies once and for all

### LINKED ISSUE #284: vulcanize alters script execution order unexpectedly
vulcanize (1.12.3) unexpectedly alters the order of script execution if an imported document has a script in the `<body>`. Example: http://pastebin.com/raw.php?i=1BPLc5iA

The unvulcanized `in.html`will run the first script first, but loading the output of `vulcanize in.html` will run the second script first.

Is this expected behavior, or a bug?

azakus:
All import portions, `<head>` and `<body>` will go into the body of the main
document. This is the only way to ensure that intra-import dependencies and
inter-import dependencies line up correctly.

Fixes #284

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -216,9 +216,14 @@ Vulcan.prototype = {
         // merge head and body tags for imports into main document
         var importHeadChildren = importHead.childNodes;
         var importBodyChildren = importBody.childNodes;
-        // replace link in head with head elements from import
-        this.replaceWith(head, importNodes[i], importHeadChildren);
-        // defer body children to be inlined in-order
+        // delete link in head
+        this.removeImportAndNewline(importNodes[i]);
+        // defer head and body children to be inlined in-order
+        if (importHeadChildren.length) {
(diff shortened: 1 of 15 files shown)


azakus:
PTAL @garlicnation @rictic

lib/vulcan.js
```diff
@@ -391,6 +381,27 @@ Vulcan.prototype = {
         return this.inlineCss(docObj.doc, docObj.href);
       }.bind(this));
     }
+
+    if (this.stripComments) {
```
rictic: Worth adding a test for stripComments

azakus: There is a check at https://github.com/Polymer/vulcanize/pull/293/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R598, but I made it more explicit in 28a7812

rictic:
LGTM!

azakus:
@sorvell @kevinpschaaf PTAL

PR #296, opened 2016-01-14
Make sure stripExclude has higher precedence than exclude

azakus:


azakus:
@garlicnation PTAL

garlicnation:
LGTM

PR #300, opened 2016-01-19
Make sure scripts moved to body have the right path

### LINKED ISSUE #297: Vulcanize is looking for scripts in the wrong place
I have a polymer element that includes a script, in my example it includes requirejs.

    <link rel="import" href="../polymer/polymer.html"/>
    <script src="../requirejs/require.js"></script>

With inlineScripts: true the vulcanize process now expects the require.js file relative to the location of the vulcanized file rather than the polymer file.

    ERROR finding resources/components/requirejs/require.js

I'm expecting the file to be somewhere else -> resources/external/bower/requirejs

This just showed up in 1.14.1.  We are looking into locking down our version to 1.14.0 to fix our build pipeline.

azakus:
Fixes #297

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -244,12 +244,18 @@ Vulcan.prototype = {
           if (end === -1) {
             end = flatHead.length;
           }
+          var container = dom5.constructors.fragment();
           for (var headidx = start, el; headidx < end; headidx++) {
             el = flatHead[headidx];
             // if the element is a script, move it to bodyFragment
             if (matchers.JS(el)) {
-              dom5.append(bodyFragment, el);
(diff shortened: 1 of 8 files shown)


azakus:
PTAL @garlicnation or @rictic

lib/vulcan.js
```diff
@@ -244,13 +244,19 @@ Vulcan.prototype = {
           if (end === -1) {
```
garlicnation: I'd like to see this broken out into a function along these lines: https://gist.github.com/garlicnation/e6137465ecd2ba9aa022

garlicnation:
LGTM

PR #302, opened 2016-01-21
Force a polymer style import to always go in a dom-module template

### LINKED ISSUE #301: Have vulcanize inline <link rel="import" type="css"> tags into the template instead of the dom-module.
Currently, vulcanize inlines `<link rel="import" type="css">` tags directly by swapping it with a `<style></style>`.

However, the correct place for `<style></style>` is inside the `<template>`.

Vulcanize should take the css import and inline it at the top of the `<template>`.

azakus:
Fixes #301


garlicnation:
LGTM

PR #309, opened 2016-02-23
Make sure all @license comments are maintained when flattening

### LINKED ISSUE #308: @license comments missing from imports
Because the inlining only copies the contents of `<head>` and `<body>`, comments with `@license` after the doctype will be missing in the output.

azakus:
Fixes #308

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -190,6 +190,15 @@ Vulcan.prototype = {
     return false;
   },
 
+  collectLicenseComments: function(nodes, collection) {
+    for (var i = 0, n; i < nodes.length; i++) {
+      n = nodes[i];
+      if (this.isLicenseComment(n)) {
+        collection.push(n);
+      }
+    }
+  },
+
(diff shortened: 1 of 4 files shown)


azakus:
PTAL @garlicnation or @justinfagnani

lib/vulcan.js
```diff
@@ -183,6 +183,22 @@ Vulcan.prototype = {
...
 
+  isLicenseComment: function(node) {
+    if (dom5.isCommentNode(node)) {
+      return dom5.getTextContent(node).indexOf('@license') > -1;
+    }
+    return false;
+  },
+
+  collectLicenseComments: function(nodes, collection) {
+    for (var i = 0, n; i < nodes.length; i++) {
```
justinfagnani: Could be `Array.push.apply(collection, nodes.filter(isLicenseComment));`

justinfagnani: Or always return a new array and `concat()` them in `flatten`

justinfagnani:
LGTM

PR #324, opened 2016-03-21
Update vulcan.js

garlicnation:
Look for scripts that hydrolysis inlines and retain their original path information.


azakus:
LGTM with update to package.json for new hydrolysis version

azakus:
nvm, doesn't depend on hydrolysis changes. LGTM

PR #330, opened 2016-04-05
Better preserve script order when interleaving script and imports

### LINKED ISSUE #325: Vulcanize rearranges script tags which breaks things
Files for repro: https://dl.dropboxusercontent.com/u/15217362/vulcanizebug.zip

If you vulcanize an import that looks like this:

```
<script src="before.js"></script>
<link rel="import" href="nested.html" />
<script src="after.js"></script>
```

You get output that looks like this:

```
<script src="after.js"></script>
<script src="before.js"></script>
```

Obviously since after.js depends on things defined in before.js, it breaks. So vulcanize ends up mangling the order of scripts and breaks things.

azakus:
Fixes #325

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -190,7 +190,7 @@ Vulcan.prototype = {
     return false;
   },
 
-  flatten: function flatten(tree, bodyFragment, mainDocUrl) {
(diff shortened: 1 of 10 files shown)


garlicnation:
LGTM

PR #339, opened 2016-04-19
Fix timeouts by making inner loop faster

### LINKED ISSUE #336: Running v1.14.11 never completes
I'm seeing this behavior on OS X 10.11.4 and Node v5.9.1.

When using the CLI v1.14.8 (the previous version published to `npm`), everything worked as expected. Using the just-released v1.14.11, the `vulcanize` command run against the same source HTML never completes, and as per the `top` command, `node` ends up pegging one my CPUs at 100% indefinitely.

I'm unfortunately not sure what `vulcanize` is doing, but if there's any specific debugging steps you'd like me to take, I'd be happy to do that. The source HTML that I'm vulcanizing is part of a private project, but I can pass that along out of band if you're not able to reproduce.

azakus:
Remove expensive `nodeWalkPrior` operation from the flatten loop, reduces timing back to v1.14.8 levels.
Remove extra whitespace when inlining, similar to old, complicated loop.
Drop jshint for eslint, fixed a few issues in the tests.

Fixes #336

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+test/html/*
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,9 @@
+{
+  "extends": "eslint:recommended",
+  "rules": {
+    "no-console": 0
+  },
+  "env": {
+    "node": true
+  }
+}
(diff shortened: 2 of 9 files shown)


azakus:
@justinfagnani, @garlicnation PTAL

garlicnation:
LGTM

PR #349, opened 2016-06-15
remove update-notifier dependency and support

FredKSchott:
It looks like this feature has been lost in time. No longer documented and opt-in only. Safe to remove support?

/cc @justinfagnani @azakus

### DIFF
--- a/bin/vulcanize
+++ b/bin/vulcanize
@@ -112,16 +112,6 @@ if (args.help || !target) {
   process.exit(0);
 }
 
-if (args['update-notifier'] !== false) {
-  (function() {
-    try {
-      require('update-notifier')({
-        pkg: pkg
-      }).notify();
-    } catch(_) {}
-  })();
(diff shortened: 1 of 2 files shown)


azakus:
It's actually opt-out, but 100% safe to remove.

FredKSchott:
Oh doy, you're right it is opt-out. We can see if anyone asks for it back :)

azakus:
Probably not, since it is really just a little annoying printout every couple of runs after a new version.

justinfagnani:
agreed, but I want to keep this in the CLI. The other CLIs not so much.

FredKSchott:
I kept it in `web-component-tester`, just because other people seem to still be using it directly.

... Which brought up another issue, which is that web-component-tester will print its own "update version"  messages even when called through polymer-cli, since we are calling the WCT CLI interface directly. 😩

PR #354, opened 2016-07-19
Do not inline imports in templates

### LINKED ISSUE #353: Imports inside of templates are incorrectly processed
Given a construction of
```html
<template>
<link rel="import" href="foo.html">
</template>
```
Vulcanize will incorrectly inline or remove `foo.html`.

azakus:
Fixes #353

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1073,12 +1073,18 @@ suite('Vulcan', function() {
           preds.hasAttrValue('rel', 'import'),
           preds.hasAttr('href')
         );
+        var externalScriptMatcher = preds.AND(
+          preds.hasTagName('script'),
+          preds.hasAttrValue('src', 'external/external.js')
+        );
         if (err) {
           return done(err);
         }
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -1065,5 +1065,22 @@ suite('Vulcan', function() {
...
+          preds.hasTagName('link'),
+          preds.hasAttrValue('rel', 'import'),
+          preds.hasAttr('href')
+        );
+        if (err) {
+          return done(err);
+        }
+        assert(doc);
+        var imports = dom5.queryAll(doc, importMatcher);
+        assert.equal(imports.length, 1, 'import in template was inlined');
```
garlicnation: Add a negative assert for the absence of one of the elements from the inlined import.

garlicnation:
LGTM with nit.

test/test.js
azakus: Done!

PR #360, opened 2016-08-16
First loop - move imports from head into body.

garlicnation:


src/test/test.ts
```diff
@@ -135,181 +130,166 @@ suite('Path Resolver', function() {
...
+      `<html><head>
+      <link rel="import" href="my-element/polymer/polymer.html">
+      <link rel="stylesheet" href="my-element/zork/my-element.css">
+      </head><body><dom-module id="my-element" assetpath="my-element/zork/">
+      <template>
+      <style>:host { background-image: url("my-element/zork/background.svg"); }</style>
+      </template>
+      </dom-module>
+      <script>Polymer({is: "my-element"})</script></body></html>`
     ].join('\n');
```
usergenic: Shouldn't this be more like:

```
    var expectedBase = `
      <html><head>
      <link rel="import" href="my-element/polymer/polymer.html">
      <link rel="stylesheet" href="my-element/zork/my-element.css">
      </head><body><dom-module id="my-element" assetpath="my-element/zork/">
      <template>
      <style>:host { background-image: url("my-element/zork/background.svg"); }</style>
      </template>
      </dom-module>
      <script>Polymer({is: "my-element"})</script></body></html>`

usergenic: (also I don't know if we care about the extra leading space now but maybe our test could have a util function to trimline like function trimlines(string) { string.replace(/^\s+/gm, '').trim() }; that we use on these heredoc style templates to ensure the leading indent doesn't leak semantic whitespace (text content nodes etc) into the test fixtures.

usergenic:
Meh- still feel like the rename of test -> testdata is unnecessary.  Other than that, LGTM.

usergenic:
Do tests need to pass to merge for this branch?  (I want to be clear on process, since I'm 'the new guy').

garlicnation:
@usergenic I'm merging this into v3, not master. I don't expect the tests to be passing for the first few PRs, since there's too much functionality to capture in one change.

garlicnation:
@usergenic The rename was unnecessary. I vastly simplified the test entry by switching back to `test/`

PR #365, opened 2016-08-29
v3 checkpoint PR

usergenic:
A checkpoint for v3 work.  Inlining HTML support, converted all tests to the promise style, added source maps support for tests, and other minor fixes.  (27 tests passing, 21 failing, 7 pending)


src/bundler.ts
```diff
@@ -125,12 +125,41 @@ class Bundler {
...
     const hidden = dom5.constructors.element('div');
     dom5.setAttribute(hidden, 'hidden', '');
     dom5.setAttribute(hidden, 'by-vulcanize', '');
-    this.removeElementAndNewline(node, hidden);
-    dom5.append(hidden, node);
+    return hidden;
+  }
+
+  /**
+   * Replace htmlImport
```
garlicnation: Fix this comment.

src/bundler.ts
```diff
@@ -125,12 +125,41 @@ class Bundler {
...
+    const resolved = url.resolve(documentUrl, rawUrl);
+    const backingScript: ScannedDocument =
+        await analyzer._analyzeResolved(resolved);
+    const scriptContent = backingScript.document.contents;
+    dom5.removeAttribute(externalScript, 'src');
+    dom5.setTextContent(externalScript, scriptContent);
+  }
+
+  /**
+   * Replace htmlImport
```
garlicnation: Fix this comment.

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
         await analyzer._analyzeResolved(resolved);
     const documentAst = dom5.parseFragment(backingDocument.document.contents);
     this.pathResolver.resolvePaths(documentAst, resolved, documentUrl);
-    const importParent = htmlImport.parentNode;
+    let importParent;
+    if (matchers.afterHiddenDiv(htmlImport)) {
```
garlicnation: This function feels too long. I'd like to see the hidden div logic split out to make it clearer what is pure inlining vs our hacks.

Also, splitting out the hidden div functionality should enable some highly needed unit tests around it.

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
...
     const body = dom5.query(newDocument, matchers.body);
-    const head = dom5.query(newDocument, matchers.head);
+    const hiddenDiv = this.getHiddenNode();
+
     const getNextImport = () => dom5.query(newDocument, matchers.htmlImport);
     const elementInHead = dom5.predicates.parentMatches(matchers.head);
     const inlinedImports = new Set<string>();
-    let c = 1;
+
     for (let nextImport; nextImport = getNextImport();) {
```
garlicnation: If you change this to a while loop, we can get tslint to stop complaining.

src/matchers.ts
```diff
@@ -16,46 +16,73 @@
...
+
+export const hiddenDiv = predicates.AND(
+    predicates.hasTagName('div'), predicates.hasAttr('hidden'),
+    predicates.hasAttr('by-vulcanize'));
+
+export const inHiddenDiv = predicates.parentMatches(hiddenDiv);
+const nextToHiddenDiv = (offset) => {
+  return (node) => {
+    const siblings = node.parentNode.childNodes;
+    const hiddenDivIndex = siblings.indexOf(node) + offset;
```
garlicnation: This matcher needs to deal with a number of edge cases. Whitespace-only text nodes should be ignored, text nodes with meaningful space should be preserved. Comments should be ignored if --strip-comments is set. License comments should be deduplicated and moved to the start of the document.

src/test/test.ts
```diff
@@ -253,309 +255,277 @@ suite('Vulcan', function() {
...
-      process(inputPath)
+      bundle(inputPath)
           .then((doc) => {
             assert.equal(dom5.queryAll(doc, imports).length, 0);
             done();
           })
-          .catch((error) => {
-            done(error);
-          });
+          .catch(done);
```
garlicnation: Does this successfully report errors? I could have sworn I tried this pattern and it didn't work, but it might have had to do with an incomplete refactor of `bundle(`

usergenic: Yeah, there's no magic.  Verified that output is equivalent to write .catch((e) => done(e));

src/bundler.ts
usergenic: putting a TODO here for inlineImportHtml refactoring branch.

src/ast-utils.ts
```diff
@@ -24,7 +24,7 @@ export default class ASTUtils {
     }
   }
 
-  static prependMultiple(target, nodes: ASTNode[]) {
+  private static prependMultiple(target, nodes: ASTNode[]) {
```
justinfagnani: style: all private names should start with `_` so that it's apparent that it's private-ish after compilation.

src/ast-utils.ts
```diff
@@ -34,6 +34,15 @@ export default class ASTUtils {
     }
   }
 
+  static insertAllBefore(target, oldNode: ASTNode, nodes: ASTNode[]) {
```
justinfagnani: why is this a static method? Can these just be exported functions?

src/bundler.ts
```diff
@@ -125,12 +125,41 @@ class Bundler {
     return false;
   }
 
-  hide(node) {
+  getHiddenNode(): ASTNode {
```
justinfagnani: comment?

src/ast-utils.ts
```diff
@@ -34,6 +34,15 @@ export default class ASTUtils {
...
+    let lastNode = oldNode;
+    for (let n = nodes.length - 1; n >= 0; n--) {
+      let node = nodes[n];
```
justinfagnani: const

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
         await analyzer._analyzeResolved(resolved);
     const documentAst = dom5.parseFragment(backingDocument.document.contents);
     this.pathResolver.resolvePaths(documentAst, resolved, documentUrl);
-    const importParent = htmlImport.parentNode;
+    let importParent;
```
justinfagnani: add type annotation

src/bundler.ts
justinfagnani: +1

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
...
+      importParent = dom5.nodeWalkPrior(htmlImport, matchers.hiddenDiv);
+      dom5.remove(htmlImport);
+      dom5.append(importParent, htmlImport);
+    } else if (matchers.beforeHiddenDiv(htmlImport)) {
+      const index = htmlImport.parentNode.childNodes.indexOf(htmlImport);
+      importParent = htmlImport.parentNode.childNodes[index + 1];
+      dom5.remove(htmlImport);
+      ASTUtils.prepend(importParent, htmlImport);
+    } else if (!matchers.inHiddenDiv(htmlImport)) {
+      let hiddenDiv = this.getHiddenNode();
```
justinfagnani: const

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
...
+    } else {
+      importParent = htmlImport.parentNode;
+    }
+
+    ASTUtils.insertAllBefore(importParent, htmlImport, documentAst.childNodes);
     dom5.remove(htmlImport);
-    ASTUtils.prependMultiple(importParent, documentAst.childNodes);
   }
 
   async bundle(url: string): Promise<ASTNode> {
```
justinfagnani: add comment

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
...
     const analyzer: Analyzer = new Analyzer(this.opts);
-    const analyzed: Document = await analyzer.analyzeRoot(url);
-    const newDocument = dom5.parse(analyzed.parsedDocument.contents);
+    const analyzedRoot: Document = await analyzer.analyzeRoot(url);
+    const newDocument = dom5.parse(analyzedRoot.parsedDocument.contents);
```
justinfagnani: did you try cloning the already parsed AST?

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
...
+    ASTUtils.insertAllBefore(importParent, htmlImport, documentAst.childNodes);
     dom5.remove(htmlImport);
-    ASTUtils.prependMultiple(importParent, documentAst.childNodes);
   }
 
   async bundle(url: string): Promise<ASTNode> {
     const analyzer: Analyzer = new Analyzer(this.opts);
-    const analyzed: Document = await analyzer.analyzeRoot(url);
-    const newDocument = dom5.parse(analyzed.parsedDocument.contents);
+    const analyzedRoot: Document = await analyzer.analyzeRoot(url);
```
justinfagnani: you shouldn't need the type annotations on these consts

src/bundler.ts
usergenic: Done.

src/bundler.ts
```diff
@@ -150,31 +179,64 @@ class Bundler {
...
-    let c = 1;
+
     for (let nextImport; nextImport = getNextImport();) {
       // If the import is in head, move all subsequent nodes to body.
       if (elementInHead(nextImport)) {
+        // Put the hiddenDiv in the body the first time we need it.
+        if (!hiddenDiv.parentNode) {
+          ASTUtils.prepend(body, hiddenDiv);
+        }
         // This function needs a better name.
```
justinfagnani: convert to a TODO

src/matchers.ts
usergenic: I'm working off the assumption that tests exist for edge cases you mentioned.  Adding TODO to ensure we have have them.

src/test/test.ts
justinfagnani: You should remove the done parameter and return the promise.

src/test/test.ts
```diff
@@ -253,309 +255,277 @@ suite('Vulcan', function() {
...
-        if (err) {
-          return done(err);
-        }
-        const svg = dom5.query(doc, preds.hasTagName('svg'));
-        assert.equal(svg.childNodes.filter(dom5.isElement).length, 6);
-        done();
-      });
-    });
+  test('svg is nested correctly', function(done) {
+    bundle(inputPath)
```
justinfagnani: Remove done argument and return the promise

src/bundler.ts
usergenic: Done.

src/bundler.ts
usergenic: Done.

src/ast-utils.ts
usergenic: SGTM.  I'll change it.

src/ast-utils.ts
usergenic: Done.

src/bundler.ts
usergenic: Done.

src/bundler.ts
usergenic: Done. Added a TODO for renaming method as well to something more intention-revealing.

src/ast-utils.ts
usergenic: Done.

src/test/test.ts
usergenic: :derp:

Well that's like a million times better.

src/bundler.ts
```diff
@@ -147,33 +183,69 @@ class Bundler {
...
+      importParent = htmlImport.parentNode;
+    }
+
+    ASTUtils.insertAllBefore(importParent, htmlImport, documentAst.childNodes);
     dom5.remove(htmlImport);
-    ASTUtils.prependMultiple(importParent, documentAst.childNodes);
   }
 
   async bundle(url: string): Promise<ASTNode> {
     const analyzer: Analyzer = new Analyzer(this.opts);
```
justinfagnani: We want the Bundler to take an Analyzer, probably in the constructor, rather than construct it's own.

src/bundler.ts
```diff
@@ -30,7 +30,7 @@ import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
 import constants from './constants';
 import * as matchers from './matchers';
 import PathResolver from './pathresolver';
-import ASTUtils from './ast-utils';
+import * as ASTUtils from './ast-utils';
 
 
 function buildLoader(config: any) {
```
justinfagnani: can this be removed?

src/bundler.ts
```diff
@@ -55,7 +55,7 @@ class Bundler {
                     String(opts.abspath).trim() !== '') ?
         path.resolve(opts.abspath) :
         null;
-    this.pathResolver = new PathResolver(this.abspath);
+    this.pathResolver = new PathResolver(Boolean(this.abspath));
```
justinfagnani: is pathResolver still used?

src/ast-utils.ts
```diff
@@ -15,32 +15,39 @@
...
 import {ASTNode} from 'parse5';
 
-export default class ASTUtils {
-  static prepend(parent, node) {
-    if (parent.childNodes && parent.childNodes.length) {
-      dom5.insertBefore(parent, parent.childNodes[0], node);
-    } else {
-      dom5.append(parent, node);
-    }
+export function prepend(parent, node) {
```
justinfagnani: Add type annotation to parameters.

justinfagnani: Is `noImplcitAny` set to `true` in tsconfig?

src/ast-utils.ts
```diff
@@ -15,32 +15,39 @@
...
 
-  static prependMultiple(target, nodes: ASTNode[]) {
-    let moveIndex = nodes.length - 1;
-    while (moveIndex >= 0) {
-      const nodeToMove = nodes[moveIndex];
-      dom5.remove(nodeToMove);
-      ASTUtils.prepend(target, nodeToMove);
-      moveIndex--;
-    }
+function _prependMultiple(target, nodes: ASTNode[]) {
```
justinfagnani: type annotation for `target` here and in `insertAllBefore`

src/bundler.ts
```diff
@@ -125,18 +125,54 @@ class Bundler {
...
     const hidden = dom5.constructors.element('div');
     dom5.setAttribute(hidden, 'hidden', '');
     dom5.setAttribute(hidden, 'by-vulcanize', '');
-    this.removeElementAndNewline(node, hidden);
-    dom5.append(hidden, node);
+    return hidden;
+  }
+
+  /**
+   * Inline external scripts <script src="*">
```
justinfagnani: Where is this script inlined? From the parameter names, I'm having trouble which ones might be mutated.

src/bundler.ts
```diff
@@ -125,18 +125,54 @@ class Bundler {
...
+   */
+  async inlineScript(
+      documentUrl: string, externalScript: ASTNode,
+      analyzer: Analyzer): Promise<void> {
```
justinfagnani: If Bundler owns an Analyzer instance, it would need to be passed in here.

src/bundler.ts
```diff
@@ -147,33 +183,69 @@ class Bundler {
     }
     inlined.add(resolved);
     const backingDocument: ScannedDocument =
-        await analyzer._analyzeResolved(resolved);
+        await analyzer._scanResolved(resolved);
     const documentAst = dom5.parseFragment(backingDocument.document.contents);
     this.pathResolver.resolvePaths(documentAst, resolved, documentUrl);
-    const importParent = htmlImport.parentNode;
+    let importParent: ASTNode;
+    if (matchers.afterHiddenDiv(htmlImport)) {
```
justinfagnani: Can you add some comments to explain the significance of checking of the import is before or after a hidden div? Maybe break this out into a `getImportParent` method too.

src/bundler.ts
justinfagnani: I'm curious how `bundle()` relates to the options given to the constructor. Is it expected that `bundle` will only be called on one particular URL, since the constructor has seemlingly document specific things, like excludes?

src/bundler.ts
```diff
@@ -147,33 +183,69 @@ class Bundler {
...
+    const newDocument = dom5.parse(analyzedRoot.parsedDocument.contents);
+
+    // Create a hidden div to target.
     const body = dom5.query(newDocument, matchers.body);
-    const head = dom5.query(newDocument, matchers.head);
-    const getNextImport = () => dom5.query(newDocument, matchers.htmlImport);
+    const hiddenDiv = this.getHiddenNode();
+
+    const getNextHtmlImport = () =>
+        dom5.query(newDocument, matchers.htmlImport);
```
justinfagnani: The scanned document should already have the list of imports, though I think we just removed the reference from Feature->Node. @rictic we may want that back.

src/bundler.ts
```diff
@@ -147,33 +183,69 @@ class Bundler {
...
     const elementInHead = dom5.predicates.parentMatches(matchers.head);
-    const inlinedImports = new Set<string>();
-    let c = 1;
-    for (let nextImport; nextImport = getNextImport();) {
+    const inlinedHtmlImports = new Set<string>();
+
+    let nextHtmlImport;
+    while (nextHtmlImport = getNextHtmlImport()) {
```
justinfagnani: It seems like there are potentially multiple walks over the document tree. Once here, and once for script inlining. Plus there's a re-parser of the contents. I wonder if this would be better as a walk over the original AST, cloning nodes to build up a new document on the fly...

src/test/test.ts
justinfagnani: 🎉

src/bundler.ts
rictic: Arg. So the main reason I was so happy about getting rid of it is that it makes serializing any features problematic. Particularly parse5's html nodes, as they have parent references.

usergenic: @rictic when you say getting rid of "it" are you referring to the hidden div?

src/bundler.ts
usergenic: Done

src/test/test.ts
usergenic: All "done" args removed from tests.  All tests return promise now.

src/ast-utils.ts
usergenic: _Done.

src/bundler.ts
usergenic: Done.

src/bundler.ts
usergenic: I've documented how I expect bundle will be called.  Bundler's configuration does appear to have lots of context-specific settings such as excludes etc.  Perhaps these will migrate into bundling options instead of Bundler options.  Will revisit this choice once tests passing.

src/bundler.ts
usergenic: Added a TODO above the newDocument assignment to look at either cloning or building it node-by-node.

src/bundler.ts
usergenic: Done

src/bundler.ts
usergenic: I put that suggestion in a TODO.  I think building up newDocument node by node is compelling.

src/ast-utils.ts
usergenic: noImplicitAny was set to false so I set it to true and also strictNullChecks to true.  And then I type'd all the things.

src/ast-utils.ts
```diff
@@ -15,32 +15,40 @@
...
+  let lastNode = oldNode;
+  for (let n = nodes.length - 1; n >= 0; n--) {
+    const node = nodes[n];
+    dom5.insertBefore(target, lastNode, node);
+    lastNode = node;
   }
 }
+
+/**
+ * Move node and its subsequent siblings to target.
```
justinfagnani: where in the target does it move them? it the target a container, or are the nodes moved to be right after it?

src/bundler.ts
```diff
@@ -30,32 +30,35 @@ import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
...
+  redirects?: string[];
+  stripComments?: boolean;
+  stripExcludes?: string[];
 }
 
+
 class Bundler {
-  constructor(opts: any) {
+  constructor(opts: Options) {
+    this.analyzer = opts.analyzer!;
```
justinfagnani: I'm not sure the `!` is correct here, as long as `analyzer` is optional in `Options`

src/bundler.ts
```diff
@@ -30,32 +30,35 @@ import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
...
     // implicitStrip should be true by default
     this.implicitStrip =
         opts.implicitStrip === undefined ? true : Boolean(opts.implicitStrip);
     this.abspath = (String(opts.abspath) === opts.abspath &&
                     String(opts.abspath).trim() !== '') ?
         path.resolve(opts.abspath) :
-        null;
-    this.pathResolver = new PathResolver(this.abspath);
+        undefined;
```
justinfagnani: Is this how clang is formatting a ternary? I'd rather see the `?` and ':' at the beginning of lines. Also, the `String()` boxing calls seem weird to me, isn't this just checking if the string is null or empty?

```
    this.abspath = (opts.abspath == null || opts.abspath.trim() == '')
      ? undefined
      : path.resolve(opts.abspath);
```

Also, also, `path.resolve(p)` implicitly depends on the CWD. I don't even know what abspath does, but if we're receiving a configured `Analyzer` is it necessary?

src/bundler.ts
```diff
@@ -30,32 +30,35 @@ import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
...
-  let loader = new FSUrlLoader(root);
-  // TODO(garlicnation): Add noopResolver for external urls.
-  // TODO(garlicnation): Add redirectResolver for fakeprotocol:// urls
-  // TODO(garlicnation): Add noopResolver for excluded urls.
-  return loader;
+import * as ASTUtils from './ast-utils';
+
+
+export interface Options {
+  abspath?: string;
```
justinfagnani: maybe not this second, but we should document these...

src/bundler.ts
```diff
@@ -66,28 +69,23 @@ class Bundler {
...
-    this.fsResolver = opts.fsResolver;
     this.redirects = Array.isArray(opts.redirects) ? opts.redirects : [];
-    this.opts = {
-      urlLoader: new FSUrlLoader(opts.root || process.cwd()),
-    };
   }
-  implicitStrip: Boolean;
-  abspath;
+
+  abspath?: string;
```
justinfagnani: style: put fields above the constructor

src/bundler.ts
```diff
@@ -118,62 +116,138 @@ class Bundler {
...
+    }
+
+    ASTUtils.insertAllBefore(importParent, htmlImport, documentAst.childNodes!);
     dom5.remove(htmlImport);
-    ASTUtils.prependMultiple(importParent, documentAst.childNodes);
   }
 
+  /**
+   * Given a URL to an entry-point html document, produce a single document
+   * suitable for deployment.
```
justinfagnani: "suitable for deployment" describes what the result might be used for, but not what the method does. "with HTML imports, external stylesheets and external scripts inlined, according to the options for this Bundler"?

src/bundler.ts
```diff
@@ -118,62 +116,138 @@ class Bundler {
...
+    // in parsedDocument.  Try cloning or even building newDocument as analyzed
+    // document is walked.
+    const newDocument = dom5.parse(analyzedRoot.parsedDocument.contents);
+
+    // Create a hidden div to target.
+    const body: ASTNode = dom5.query(newDocument, matchers.body)!;
+    const hiddenDiv = this.getHiddenNode();
+
+    const getNextHtmlImport = () =>
+        dom5.query(newDocument, matchers.htmlImport);
```
justinfagnani: this is going to query each time its called. Would it be correct to queryAll once and iterate through the list?

src/bundler.ts
```diff
@@ -30,32 +30,35 @@ import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
...
-  const excludes = config.excludes;
-  const fsResolver = config.fsResolver;
-  const redirects = config.redirects;
-  let root = abspath && path.resolve(abspath) || process.cwd();
-  let loader = new FSUrlLoader(root);
-  // TODO(garlicnation): Add noopResolver for external urls.
-  // TODO(garlicnation): Add redirectResolver for fakeprotocol:// urls
-  // TODO(garlicnation): Add noopResolver for excluded urls.
-  return loader;
+import * as ASTUtils from './ast-utils';
```
justinfagnani: style: name the library lowercase so it doesn't look like static methods at call sites.

src/bundler.ts
```diff
@@ -118,62 +116,138 @@ class Bundler {
...
     const backingDocument: ScannedDocument =
-        await analyzer._analyzeResolved(resolved);
+        await this.analyzer._scanResolved(resolved);
     const documentAst = dom5.parseFragment(backingDocument.document.contents);
     this.pathResolver.resolvePaths(documentAst, resolved, documentUrl);
-    const importParent = htmlImport.parentNode;
+    let importParent: ASTNode;
+    if (matchers.afterHiddenDiv(htmlImport)) {
+      importParent = dom5.nodeWalkPrior(htmlImport, matchers.hiddenDiv)!;
+      dom5.remove(htmlImport);
```
justinfagnani: TODO: remove the `remove()` call when https://github.com/PolymerLabs/dom5/issues/35 is fixed

src/bundler.ts
```diff
@@ -118,62 +116,138 @@ class Bundler {
...
+      dom5.append(importParent, htmlImport);
+    } else if (matchers.beforeHiddenDiv(htmlImport)) {
+      const index = htmlImport.parentNode!.childNodes!.indexOf(htmlImport);
+      importParent = htmlImport.parentNode!.childNodes![index + 1];
+      dom5.remove(htmlImport);
+      ASTUtils.prepend(importParent, htmlImport);
+    } else if (!matchers.inHiddenDiv(htmlImport)) {
+      const hiddenDiv = this.getHiddenNode();
+      dom5.replace(htmlImport, hiddenDiv);
+      dom5.append(hiddenDiv, htmlImport);
```
justinfagnani: Is this correct? Don't you want to append the import's document contents?

src/pathresolver.ts
```diff
@@ -28,21 +28,21 @@ class PathResolver {
   constructor(public abspath?: boolean) {
   }
 
-  isTemplatedUrl(href) {
+  isTemplatedUrl(href: string): boolean {
```
justinfagnani: comment?

justinfagnani:
@usergenic can you update .travis.yml to only run tests on node 4 and 6?

src/bundler.ts
usergenic: That whole segment of code was cargo'd in from the old vulcan.js.  It could be reworked to something with clearer intent.  I'll fix it up.  The String() bit does seem unnecessary.

The ternary formatting is clang's choice.

I'm not 100% sure what abspath is for.  I've not fully come to understand or encounter the way they are being exercised yet in tests.  I'll be documenting and/or removing each option fully in the Options type for bundler soon.

usergenic: There is a big ugly PathResolver module cargo'd in from old vulcanize that probably needs to be refactored to take analyzer's resolver into account, but it apparently uses abspath.  Will replace or update it once I have all the tests covering path resolution working.

src/bundler.ts
usergenic: yeah definitely going to document them...once i know what they're for.

src/bundler.ts
usergenic: Done.

src/bundler.ts
usergenic: Agreed.  Thanks for writing that for me :)

src/bundler.ts
usergenic: That's a very good point.  I think perhaps this approach must have been a lazy way to ensure nested imports worked; definitely need to fix this.  Adding a TODO.

src/bundler.ts
usergenic: Ok.  Done.

src/bundler.ts
usergenic: Added.

src/pathresolver.ts
usergenic: PathResolver is going to undergo some refactoring.  I only added type annotations here to existing stuff because lint warnings.  Will add comments when refactored to support new root+abspath behavior.

PR #369, opened 2016-09-08
Refactoring related to abspath behavior in path resolver and various cleanups in bundler and test fixtures.

usergenic:
This contains several items cleaning up code that was intended to be merged in the big PR from inlining branch, but items were missed.

There's initial work related to abspath behavior in the path resolver, but some further discussion about desired behavior of abspath as it relates to prefixing existing absolute urls etc, needs to be hashed out before that is implemented.

Also there was semantic confusion in test fixtures for the word "external" as it seems to have meant two different scenarios "content not already inlined" and "remote unanalyzed imports".  I tried to reconsile this by renaming the remote imports from "external-*" to "remote-*".


usergenic:
Apologies for having these concerns intertwingled in the same PR.

src/pathresolver.ts
```diff
@@ -84,7 +87,7 @@ class PathResolver {
     }
     const absUrl = url.resolve(importUrl, relUrl);
     if (this.abspath) {
-      return url.resolve('/', absUrl);
+      return url.resolve(this.abspath, absUrl);
```
garlicnation: This should all happen after all the inlining, or it will get confusing to resolve in cases where the abspath contains a protocol and domain.

src/bundler.ts
```diff
@@ -267,12 +282,13 @@ class Bundler {
     // should be removed from the document.
     const importMap: Map<string, Import|null> = new Map();
     analyzedRoot.getByKind('import').forEach((i) => importMap.set(i.url, i));
-    this.excludes.forEach((u) => {
-      if (importMap.has(u)) {
+    importMap.forEach((_, u) => {
```
garlicnation: This is n^2, but I'm not sure how to avoid it if we consider excludes by suffix.

garlicnation: It might make more sense to only allow excludes specified as root-relative paths, which would let us do O(1) lookups.

FredKSchott:
looking good, @garlicnation touched on everything afaict

src/bundler.ts
```diff
@@ -79,7 +79,7 @@ class Bundler {
                     String(opts.abspath).trim() !== '') ?
         path.resolve(opts.abspath) :
         undefined;
-    this.pathResolver = new PathResolver(Boolean(this.abspath));
+    this.pathResolver = new PathResolver(this.abspath);
```
usergenic: tl;dr - I think there are two separate but related features, both optional and I propose the names `basePath` and `absPathPrefix` for them.  The filesystem`root` is not provided or used in this context.

---

When we discussed `root` before, it was in the context of the filesystem root (at least that was my understanding) when telling bundler to go bundle.

The thing is that the filesystem's folder root should make no difference to this PathResolver (maybe PathResolver should be renamed to avoid confusion with Analyzer's resolvers), precisely because they have jurisdiction over and may totally redefine relationship of physical files to everything else, where PathResolver honors the resolved url topology.

Besides which, there are no opportunities to apply the filesystem `root` folder concept in PathResolver logic.

However, there might be reason to support two arguments:
 - a path to resolve relative urls to.  typically '/' but could be '/myapp/' etc.
 - a prefix to apply to absolute urls.

The value of separating those options would be to allow expression of absolute hrefs for assets known to be peer-provided in a deployment that would be unaffected by the specific mounting root of the application `/path/to/myapp`.

We could call them `root` and `prefix` but I don't want `root` to be confused with `root` as we use for targeting files to bundle...  Also, applying a prefix to existing absolute paths is a new/separate/different bundler feature from resolving relative paths to an absolute form AFAIK.

Perhaps `basePath` would be a more appropriate term than `root`?  In the sense that `basePath`  applies only to relative paths only whereas having no effect on absolute paths.  Then `absPathPrefix` could be the term used to express the prefix applied to ALL absolute paths.

In this way they could also function independently.

usergenic: Perhaps PathRewriter is a better name than PathResolver

src/pathresolver.ts
```diff
@@ -25,7 +25,10 @@ import constants from './constants';
 const pathPosix = path.posix;
 
 class PathResolver {
```
justinfagnani: Can we get some docs on this class and it's methods?

src/bundler.ts
garlicnation: I think we can get away with just a `prefix`, which could be applied by a final pass of the pathResolver after all inlining is completed. I think there's a value to the `basePath` argument, but I'm not sure that we've had any demand for that.

The concept of a `root` will have to exist as a command line argument, but nothing in the Bundler class needs to know about it, since the Analyzer will take care of it.

garlicnation: +1 to PathRewriter.

I'd even be up for a more verbose name like UrlAttributeRewriter.

justinfagnani: What's the use case for the first option? It seems the same as `<base>`, but I'm not sure when we'd want to only affect relative URLs, not all URLs.

URLResolvers, as you note, have nearly complete control over the URL space, but implicitly (and we should make this more clear if necessary), it's expected that all URLs are resolved to the base URL of their enclosing document first, making them effectively all absolute anyway.

We definitely need to be able to prefix all of the absolute URLs to enable mounting an app at a subfolder. We've had this requested a bunch (I can't find the issue in polymer-cli or polymer-build, @FredKSchott ?). I'm not sure that bundler is the place for it, as opposed to polymer-build. In fact, it seems like the only path munging that the bundler should be doing is the minimal necessary because it's moving content to different document URLs.

If that's not the only thing that PathResolver does, what else is it doing and can we get rid of it?

src/pathrewriter.ts
```diff
@@ -24,15 +24,18 @@ import constants from './constants';
 
 const pathPosix = path.posix;
 
-class PathResolver {
-  constructor(public abspath?: boolean) {
+class PathRewriter {
```
justinfagnani: I left this comment on another commit I guess, but I hate how github shows these things: Can we get comments on this class and it's methods?

justinfagnani: This class has no real state outside `abspath`, which is only used in `rewriteRelPath`. It should probably just be a module of functions.

src/bundler.ts
usergenic: I'm totally missing the O(n^2) -- This is an O(n) map population.  Maybe I am wearing blinders somehow?

src/pathrewriter.ts
usergenic: Moved to module of functions called urlUtils.  The actual ast rewriter function is moved to Bundler.

src/pathresolver.ts
usergenic: The class is now gone.  I have added documentation for the exported functions of url-utils.ts (which is where most of the code lives).

src/bundler.ts
```diff
@@ -136,7 +163,7 @@ class Bundler {
    * appended.
    * TODO(usergenic): Give this a more intention-revealing name.
```
justinfagnani: is this done now? :)

src/bundler.ts
```diff
@@ -191,14 +218,20 @@ class Bundler {
 
     const imprt = importMap.get(resolvedUrl);
     if (imprt) {
+      // If the document wasn't loaded for the import during analysis, we can't
+      // inline it.
+      if (!imprt.document) {
+        // TODO(usergenic): What should the behavior be when we don't have the
+        // document to inline available in the analyzer?
+        return;
```
justinfagnani: I'm very curious when this even happens. Is it an expected situation when running across external imports, say? For now maybe console a debug would be good, but eventually it'd be nice to have a more clearly meaningful condition to check on, especially so we can throw an error if a document doesn't load that we _should_ inline.

src/bundler.ts
```diff
@@ -191,14 +218,20 @@ class Bundler {
 
     const imprt = importMap.get(resolvedUrl);
     if (imprt) {
```
justinfagnani: for future reference, no action needed: where the logical name is reserved or already taken I prefer adding something like `_`, or making a long name like `analzedImport` than mangling the name.

src/bundler.ts
```diff
@@ -243,7 +276,7 @@ class Bundler {
 
   // TODO(usergenic): Migrate "Old Polymer" detection to polymer-analyzer with
   // deprecated feature scanners.
-  oldPolymerCheck(analyzedRoot: Documents) {
+  oldPolymerCheck(analyzedRoot: Document) {
```
justinfagnani: I bet we could just remove this altogether.

src/test/test.ts
```diff
@@ -231,30 +203,32 @@ suite('Path Resolver', function() {
...
 
-    const actual = serialize(ast);
+    const actual = dom5.serialize(ast);
     assert.equal(actual, base, 'templated urls');
   });
 
 });
 
-suite('Vulcan', function() {
+suite('Vulcan', () => {
```
justinfagnani: 'Vulcan' -> 'Bundler'

src/test/test.ts
```diff
@@ -766,16 +728,20 @@ suite('Vulcan', function() {
     });
   });
 
-  suite('Input URL', function() {
+  // TODO(usergenic): These tests only prove that the `inputUrl` has precedence
+  // over the filename presented to `bundle(path)`.  Do we want to continue to
+  // support inputUrl?  Tese don't prove anything about the doc production
```
justinfagnani: @garlicnation do we have use cases for inputUrl?

justinfagnani:
LGTM

src/bundler.ts
usergenic: I'm on board with that.

src/bundler.ts
usergenic: Well, I think the analyzer will need to be able to continue its work when it fails to load an import.  At least for the linter, this behavior is high value.  As for the bundler, it might be nice to be able to bundle incompletely especially in a development context.  Rather than throwing, we should consider building some kind of Errors collection we can return at the end of processing.  Of course, we can provide option to fail-fast, too.

src/bundler.ts
usergenic: I'll leave this here for this PR but we can remove after we have a version supported/unsupported story nailed down.

src/bundler.ts
justinfagnani: I don't know, since you can develop without bundling, I'm not sure that a file load shouldn't be a fatal error.

PR #370, opened 2016-09-15
Add legacy command line interface

garlicnation:


src/bin/polymer-bundler.ts
```diff
@@ -0,0 +1,178 @@
...
+  If the \`--strip-excludes\` flag is used, the HTML Import \`<link>\` tags that point to resources in \`path/totarget/subpath\` and \`path/to/target/subpath2/\` will also be removed.
+
+  The command
+    vulcanize --inline-scripts target.html
+  will inline scripts in \`target.html\` as well as HTML Imports. Exclude flags will apply to both Imports and Scripts.
+  `,
+  raw: true}
+]
+
+    const options = commandLineArgs(optionDefinitions);
```
justinfagnani: Leading spaces? Was this run though the formatter?

src/bin/polymer-bundler.ts
```diff
@@ -0,0 +1,178 @@
...
+options.redirects = options.redirect || [];
+options.stripExcludes = options['strip-exclude'] || [];
+options.stripComments = options['strip-comments'];
+options.implicitStrip = !options['no-implicit-strip'];
+options.inlineScripts = options['inline-scripts'];
+options.inlineCss = options['inline-css'];
+console.log(options);
+options.analyzer = new Analyzer({urlLoader: new FSUrlLoader()};
+
+(new Bundler(options)).bundle(target).then((content) => {
```
justinfagnani: You could use an IIAFE here (A = Async), and use await.

tsconfig.json
```diff
@@ -4,8 +4,8 @@
         "module": "commonjs",
         "moduleResolution": "node",
         "isolatedModules": false,
-        "noImplicitAny": true,
-        "strictNullChecks": true,
+        "noImplicitAny": false,
```
justinfagnani: Did you remove these for development? I'd really prefer we keep them.

tsconfig.json
```diff
@@ -15,6 +15,8 @@
         "sourceMap": true
     },
     "include": [
+        "custom_typings/*.d.ts",
+        "node_modules/**/*.d.ts",
```
justinfagnani: We shouldn't include all .d.ts file in node_modules. tsc will automatically look in `@types` and in packages that are imported (looking for the "typings" field").

src/bin/polymer-bundler.ts
garlicnation: The formatter is getting tripped up on something in this document and causing problems. I've had to revert a ton of egregiously horrible whitespace changes. I'll try and investigate what the cause is.

tsconfig.json
garlicnation: Removed - these weren't working when I opened up this branch.

tsconfig.json
garlicnation: Ack.

tsconfig.json
usergenic: Hmm...  The `noImplicitAny: true` and `strictNullChecks: true` are not causing any problems on v3 branch or inline-css branch, which was cut from it.  Where did this addCli branch get cut from?

garlicnation:
@usergenic PTAL

src/bundle-manifest.ts
```diff
@@ -174,6 +199,7 @@ export function generateShellMergeStrategy(
   return (bundles: Bundle[]): Bundle[] => {
     const newBundles = generateSharedDepsMergeStrategy(minEntrypoints)(bundles);
     const shellBundle = newBundles.find((bundle) => bundle.files.has(shell));
+    console.log(bundles);
```
usergenic: remove

src/bundler.ts
```diff
@@ -488,64 +499,115 @@ class Bundler {
...
-      bundleManifest: BundleManifest,
-      bundleImports?: Set<string>): Promise<ASTNode> {
-    const analyzedRoot = await this.analyzer.analyze(url);
-
-    // Set tracking imports that have been reached.
-    const inlinedImports: Set<UrlString> = new Set();
-
-    // We must parse document to a new AST since we will be modifying the AST.
-    const newDocument = parse5.parse(analyzedRoot.parsedDocument.contents);
+  private _computeRelativeUrl(from: UrlString, to: UrlString): string
```
usergenic: We already have this:
https://github.com/Polymer/vulcanize/blob/v3/src/url-utils.ts#L44

Which has lots of tests because tricky cases:
https://github.com/Polymer/vulcanize/blob/v3/src/test/url-utils_test.ts#L67

usergenic:
Sorry for the delay on this review!

src/bundle-manifest.ts

src/bundler.ts
```diff
@@ -278,7 +278,7 @@ class Bundler {
         dom5.remove(htmlImport);
         return;
       }
-      const relative = this._computeRelativeUrl(docUrl, bundleUrl) || bundleUrl;
+      const relative = urlUtils.relativeUrl(docUrl, bundleUrl) || bundleUrl;
```
usergenic: I should mention is that if the relativeUrl can not be calculated because of host/protocol mismatch, it returns bundleUrl.  If bundle.url == targetBundle.url here then relative will be an empty string and will throw.  Can that case occur?

Additionally, if docUrl == bundleUrl then bundleUrl will be returned in full because relative url is an empty string.

src/bundler.ts
```diff
@@ -132,7 +132,7 @@ class Bundler {
       return false;
     }
     if (targetBundle.url !== bundle.url) {
-      const relative = this._computeRelativeUrl(bundle.url, targetBundle.url);
+      const relative = urlUtils.relativeUrl(bundle.url, targetBundle.url);
```
usergenic: If bundle.url == targetBundle.url here then relative will be an empty string and will throw.  Can that case occur?

usergenic: Oh derp of course it can't because the urls are !== compared by the if.  Derp.  Ignore lol.

src/bundler.ts
garlicnation: Got rid of the `|| bundleUrl`

usergenic:
LGTM

PR #371, opened 2016-09-16
CSS Inlining

usergenic:
CSS inlining is now implemented and works nearly identically as before.  I made one change in behavior, which was to move media query from `<link media="*">` tag directly to `<style media="*">` because cleaner.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -170,11 +170,20 @@ class Bundler {
   }
 
   /**
+   * Creates a <style> tag to which inlined content will be appended.
+   */
+  createInlineStyleNode() {
+    const styleNode = dom5.constructors.element('style');
+
+    return styleNode;
+  }
+
+  /**
    * Inline external scripts <script src="*">
    */
   inlineScript(
       docUrl: string, externalScript: ASTNode,
-      importMap: Map<string, Import|null>) {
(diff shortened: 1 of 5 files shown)


garlicnation:
LGTM.

Nice!

src/bundler.ts
```diff
@@ -188,21 +186,40 @@ class Bundler {
         encodeString(script.document.parsedDocument.contents, true);
     dom5.removeAttribute(externalScript, 'src');
     dom5.setTextContent(externalScript, scriptContent);
+
+    return externalScript;
```
justinfagnani: It'd be good to have a comment that this returns the external script, and probably that it mutates some of its inputs (though this probably applies to lots of methods in the bundler...)

src/bundler.ts
```diff
@@ -188,21 +186,40 @@ class Bundler {
...
+        resolvedStylesheetUrl, docUrl, stylesheetContent);
+    const styleNode = dom5.constructors.element('style');
+
+    if (media) {
+      dom5.setAttribute(styleNode, 'media', media);
+    }
+
+    dom5.replace(cssLink, styleNode);
+    dom5.setTextContent(styleNode, resolvedStylesheetContent);
+    return styleNode;
```
justinfagnani: same here, re: comments about return value

PR #372, opened 2016-09-16
Bundle now returns a collection of documents keyed by url instead of a single AST

garlicnation:


justinfagnani:
What does it mean to "support a collection of documents"? :)

usergenic:
We know we're going to be generating multiple documents as we move sharding logic into bundler so the document set result type feels non-controversial, even if other details are still unknown.

justinfagnani:
I still don't grok "support a collection of documents" though. Support where?

justinfagnani:
Ok, looked at the code, and know I know :)

LGTM

I do want to encourage comments, PR descriptions, commit messages, method and variable names, etc., to be more descriptive and targeted at new readers :)

PR #373, opened 2016-09-19
Bundles

garlicnation:
Add basic functions for generating a depsindex, with tests.


src/bundler.ts
```diff
@@ -345,12 +349,70 @@ class Bundler {
    * Given a URL to an entry-point html document, produce a single document
    * with HTML imports, external stylesheets and external scripts inlined,
    * according to the options for this Bundler.
+   *
+   * Given Multiple urls, produces a sharded build.
```
justinfagnani: how? what determines the shards?

src/bundler.ts
```diff
@@ -79,6 +79,7 @@ export interface Options {
   stripExcludes?: string[];
 }
 
+type DependencyEntry = {url: string, eager: string[], lazy: string[]};
```
justinfagnani: what's this? It sounds singular, but looks plural. Is it a value in a map?

src/bundler.ts
```diff
@@ -345,12 +349,70 @@ class Bundler {
...
+   * Given Multiple urls, produces a sharded build.
    */
-  async bundle(url: string): Promise<DocumentCollection> {
+  async bundle(url: string|string[]): Promise<DocumentCollection> {
     if (!this.analyzer) {
       throw new Error('No analyzer provided.');
     }
-    const analyzedRoot = await this.analyzer.analyzeRoot(url);
+    if (Array.isArray(url)) {
+      return this._bundleMultiple(url, this.analyzer);
```
justinfagnani: could we just wrap the string in an array and pass it onto a single code-path?

justinfagnani:
I'd like to start with tests that look at the dependency indices. In polymer-build we have this data-structure:

```typescript
export interface DepsIndex {
  // An index of dependency -> fragments that depend on it
  depsToFragments: Map<string, string[]>;
  // TODO(garlicnation): Remove this map.
  // An index of fragments -> html dependencies
  fragmentToDeps: Map<string, string[]>;
  // A map from frament urls to html, js, and css dependencies.
  fragmentToFullDeps: Map<string, DocumentDeps>;
}
```

And we also talked about the map of `Tuple of Entrypoints -> Deps that are depended on by those`.

Once we know we can build the correct indices, we can build include/excludes lists, then wire in the bundling.

src/bundler.ts

src/bundler.ts
garlicnation: Called out the heuristic argument.

garlicnation:
@justinfagnani @usergenic PTAL

src/bundler.ts
```diff
@@ -111,6 +109,9 @@ class Bundler {
         String(opts.inputUrl) === opts.inputUrl ? opts.inputUrl : '';
   }
 
+  calculateImplicitExcludes(excludes: string[]) {
```
justinfagnani: comment or remove?

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
+    collection.set(bundles[0], doc);
+    return collection;
+    // const depsIndex = buildDepsIndex(bundles, this.analyzer) const manifest =
+    //     this._bundleManifest() return this._bundleMultiple(url,
+    //     this.analyzer);
+  }
+
+
+
+  private async _bundleMultiple(entrypoints: string[], analyzer: Analyzer):
```
justinfagnani: Looks like we keep passing an analyzer around. Could we just add it as a constructor argument?

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
+      Promise<DocumentCollection> {
+    const bundles: DocumentCollection = new Map();
+    // Map entrypoints to transitive dependencies.
+    const depsIndex = await buildDepsIndex(entrypoints, analyzer);
+
+
+    // Determine root of dependency graph, if applicable.
+    let root: string|null = null;
```
justinfagnani: factor out into `getRoot()`?

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
+          foundRoot = false;
+          break;
+        }
+      }
+      if (!foundRoot) {
+        continue;
+      }
+      if (root) {
+        throw new Error(
+            `entrypoints must form a singly-rooted graph! Two roots found: "${root
```
justinfagnani: why is this an error? we should support multiple entrypoints.

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
+            }" and "${entrypoint}"!`);
+      }
+      root = entrypoint;
+    }
+    if (!root) {
+      throw new Error(
+          'entrypoints must form a singly-rooted graph! No roots found!')
+    }
+
+    for (const entrypointUrl of entrypoints) {
```
justinfagnani: can you add a comment about what this loop is doing?

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
+      let whitelist: Set<string>;
```
justinfagnani: what's a whitelist?

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
+      if (entrypointUrl == root) {
+        whitelist = depsIndex.fragmentToDeps.get(entrypointUrl)!;
+      } else {
+        whitelist = depsIndex.fragmentToDeps.get(entrypointUrl)!;
+        for (const dependency of whitelist.values()) {
+          console.log('Considering removing: ', dependency);
```
justinfagnani: remove

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
+          if (dependency in entrypoints) {
+            const dependenciesToExclude =
```
justinfagnani: I'm confused about building up this exclusion list. Shouldn't we know exactly which dependencies are exclusive to this bundler after building the dependencies indices? We should have one index of (bundle)->[exclusive-dependencies] where [exclusive-dependencies] contains nothing in any other bundle.

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+export interface BundleManifest extends Map<string, Set<string>> {}
+
+export interface BundleOrchestrator {
+  (endpoints: string[], depsIndex: DepsIndex): BundleManifest
+}
+
+type DependencyMapEntry = {
+  url: string,
+  eager: Set<string>,
+  lazy: Set<string>
```
justinfagnani: is this a set of things only reachable lazily?

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+async function _getDependencies(href: string, analyzer: Analyzer):
```
justinfagnani: is this direct dependencies or transitive?

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+import {Analyzer} from 'polymer-analyzer';
+
+export interface DepsIndex {
+  // An index of dependency -> fragments that depend on it
+  depsToFragments: Map<string, Set<string>>;
+  // An index of fragments -> html dependencies
+  fragmentToDeps: Map<string, Set<string>>;
+}
+
+export interface BundleManifest extends Map<string, Set<string>> {}
```
justinfagnani: what do the strings represent? We've talked about a bundle as identified as a tuple of entrypoints....

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+async function _getDependencies(href: string, analyzer: Analyzer):
+    Promise<DependencyMapEntry> {
+      const document = await analyzer.analyzeRoot(href);
+      const imports = document.getByKind('import');
+      const eagerImports = new Set<string>();
+      const lazyImports = new Set<string>();
+      for (let htmlImport of imports) {
+        if (!htmlImport.url) {
+          continue;
```
justinfagnani: have you seen this? what cases produce an import feature with no URL?

src/deps-index.ts
justinfagnani: naming: `href` -> `url`

src/bundler.ts
```diff
@@ -111,6 +109,9 @@ class Bundler {
         String(opts.inputUrl) === opts.inputUrl ? opts.inputUrl : '';
   }
 
+  calculateImplicitExcludes(excludes: string[]) {
+  }
+
   isExcludedHref(href: string): boolean {
```
justinfagnani: naming: `href` -> `url`

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+      }
+      return {url: href, eager: eagerImports, lazy: lazyImports};
+    }
+
+export async function buildDepsIndex(entrypoints: string[], analyzer: Analyzer):
+    Promise<DepsIndex> {
+  const entrypointToDependencies: Map<string, Set<string>> = new Map();
+  const dependenciesToEntrypoints: Map<string, Set<string>> = new Map();
+  let depsIndex = {};
+  const queue = entrypoints.slice();
```
justinfagnani: use `Array.from()`

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+            break;
+        }
+      }
+      return {url: href, eager: eagerImports, lazy: lazyImports};
+    }
+
+export async function buildDepsIndex(entrypoints: string[], analyzer: Analyzer):
+    Promise<DepsIndex> {
+  const entrypointToDependencies: Map<string, Set<string>> = new Map();
+  const dependenciesToEntrypoints: Map<string, Set<string>> = new Map();
```
justinfagnani: can this function also create the (entrypoint)->[exclusive-dependencies] map?

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  });
+  return {
+    depsToFragments: dependenciesToEntrypoints,
+    fragmentToDeps: entrypointToDependencies
+  };
+}
+
+
+export function defaultHeuristic(entrypoints: string, index: DepsIndex):
```
justinfagnani: Let's name the built-in heuristics. I think we should have at least one-bundle and maximum sharding (the identity function on the bundle manifest).

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+    BundleManifest {
+      // Move all bundles into a shared bundle
+      return new Map<string, Set<string>>();
```
justinfagnani: how does an empty map represent a single bundle? Shouldn't it have one entry with all entrypoints in the key and all dependencies in the value?

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+export interface DepsIndex {
+  // An index of dependency -> fragments that depend on it
+  depsToFragments: Map<string, Set<string>>;
+  // An index of fragments -> html dependencies
+  fragmentToDeps: Map<string, Set<string>>;
+}
+
+export interface BundleManifest extends Map<string, Set<string>> {}
+
+export interface BundleOrchestrator {
```
usergenic: PTAL at https://github.com/Polymer/vulcanize/pull/379 where I changed the interface for the orchestrator to a BundleStrategy type that has a pass-thru noop instead of a `(endpoints, deps) => manifest` signature.

src/deps-index.ts
```diff
@@ -0,0 +1,99 @@
...
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+import {Analyzer} from 'polymer-analyzer';
+
+export interface DepsIndex {
+  // An index of dependency -> fragments that depend on it
+  depsToFragments: Map<string, Set<string>>;
+  // An index of fragments -> html dependencies
+  fragmentToDeps: Map<string, Set<string>>;
```
usergenic: The only kind of map which matters is this fragmentToDeps one since we have a function to produce a set of initial bundle definitions based purely on it.  The depsToFragments is ephemeral and only really used when performing this initial bundle set construction.

usergenic: And when I say "we have a function" I mean its in #379

src/deps-index.ts
usergenic: In #380 I defined a BundleManifest class which has the accessor for bundle lookup based on constituent file as well as URL mapping.  These will both be important as we translate imports in the bundling operation.

src/bundler.ts
garlicnation: Removed this.

src/bundler.ts
garlicnation: Removed

src/bundler.ts
garlicnation: Changed argument name, function names will change in a subsequent PR.

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
...
    */
-  async bundle(url: string): Promise<DocumentCollection> {
+  async bundle(bundles: string[], heuristic?: BundleOrchestrator):
+      Promise<DocumentCollection> {
     if (!this.analyzer) {
       throw new Error('No analyzer provided.');
     }
-    const analyzedRoot = await this.analyzer.analyzeRoot(url);
+    const doc = await this._bundleDocument(
+        bundles[0], this.excludes, this.stripExcludes, this.analyzer);
```
garlicnation: We can eliminate them, but I'd like to do that in a follow-up when we finish implementing sharding.

src/bundler.ts
garlicnation: I want to move away from the Object-based model to a function-based model. This is an intermediate step to aid testability.

src/bundler.ts
garlicnation: Removed from this PR

src/bundler.ts
garlicnation: Also removed from this PR.

src/bundler.ts
garlicnation: removed

src/bundler.ts
garlicnation: removed

src/bundler.ts
garlicnation: removed

src/deps-index.ts
garlicnation: Added a comment.

src/deps-index.ts
garlicnation: Transitive - renamed the function

src/deps-index.ts
garlicnation: I haven't seen this, but I want to keep it in for safety. Should I throw here?

src/deps-index.ts
garlicnation: That's handled in the later bundling functions in @usergenic's code. I wanted to keep this more mechanical.

src/deps-index.ts
garlicnation: Done.

src/deps-index.ts
garlicnation: This is duplicated in @usergenic's branch. Removed.

src/deps-index.ts
garlicnation: Cleaned up the map

src/bundler.ts
garlicnation: Killed this loop entirely.

justinfagnani:
Looks like you need to update the analyzer to 2.0.0-alpha.9 - the API changed

src/bundler.ts
```diff
@@ -387,11 +386,30 @@ class Bundler {
    * Given a URL to an entry-point html document, produce a single document
    * with HTML imports, external stylesheets and external scripts inlined,
    * according to the options for this Bundler.
+   *
+   * Given Multiple urls, produces a sharded build by applying the provided
+   * heuristic.
```
justinfagnani: heuristic -> orchestrator

src/deps-index.ts
justinfagnani: `console.assert` maybe

src/bundler.ts
usergenic: orchestrator -> strategy

package.json
```diff
@@ -28,7 +29,7 @@
     "typings": "^1.3.2"
   },
   "scripts": {
-    "test": "npm run format && tsc && tslint -c tslint.json src/*.ts src/**/*.ts ; mocha",
```
usergenic: clang-format branch merged-- i think this was removed because formatting was being lame-- hopefully new formatting doesn't cause problems and we can keep aggressively clang-format'n

usergenic:
v3 contains polymer-analyzer 2.0.0-alpha.7 peg in package.json which will fix the travis reported incompatibility with recent analyzer's removal of analyzeRoot.

src/deps-index.ts
garlicnation: Done.

src/deps-index.ts
garlicnation: Changed. I wrapped it in a try/catch since I still don't believe this should kill the pipeline. LMK if you think I should just let it throw a wrench into the works.

src/deps-index.ts
garlicnation: Removed this.

src/deps-index.ts
garlicnation: Rebased the code and killed this interface in favor of BundleStrategy

src/deps-index.ts
garlicnation: This PR now uses @usergenic's types from #380

src/deps-index.ts
garlicnation: Done.

src/bundler.ts
```diff
@@ -386,12 +387,93 @@ class Bundler {
    * Given a URL to an entry-point html document, produce a single document
    * with HTML imports, external stylesheets and external scripts inlined,
    * according to the options for this Bundler.
+   *
+   * Given Multiple urls, produces a sharded build.
    */
-  async bundle(url: string): Promise<DocumentCollection> {
+  async bundle(bundles: string[], heuristic?: BundleOrchestrator):
```
garlicnation: Changed to entrypoints and updated docs to match our intended behavior.

src/bundler.ts
```diff
@@ -409,7 +440,7 @@ class Bundler {
     });
 
     // We must clone the AST from the document, since we will be modifying it.
```
usergenic: Update comment or remove--?

Maybe: "We must parse document to a new AST since we will be modifying it."
Maybe also a note why cloneNode is insufficient to this task?

src/bundler.ts
```diff
@@ -482,10 +513,7 @@ class Bundler {
         }
       });
     }
-    const documents = new Map<string, ASTNode>();
-    // TODO(garlicnation): Set to the resolved document url?
-    documents.set(url, newDocument);
-    return documents;
+    return newDocument;
     // TODO(garlicnation): inline CSS
```
usergenic: Oh hey, we can remove this TODO now :)

src/bundler.ts
```diff
@@ -387,11 +395,34 @@ class Bundler {
...
+   * @param {Array<string>} entrypoints The list of entrypoints that will be
+   *     analyzed for dependencies. The results of the analysis will be passed
+   *     to the `strategy`. An array of length 1 will bypass the strategy and
+   *     directly bundle the document.
+   * @param {BundleStrategy} strategy The strategy used to construct the
+   *     output bundles. See 'polymer-analyzer/lib/bundle-manifest' for
+   *     examples. UNUSED.
    */
-  async bundle(url: string): Promise<DocumentCollection> {
+  async bundle(entrypoints: string[], strategy?: BundleStrategy):
```
usergenic: Until we support multiple entrypoints for realsies maybe a `console.assert` that its length === 1?

usergenic:
Travis reported this for node 6.  node 4 output was useless, but they're both failing.
```
src/test/deps-index_test.ts(107,39): error TS2339: Property 'depsToFragments' does not exist on type 'DepsIndex'.
src/test/deps-index_test.ts(108,39): error TS2339: Property 'fragmentToDeps' does not exist on type 'DepsIndex'.
```

src/bundler.ts
garlicnation: Done.

src/bundler.ts
garlicnation: Done.

src/bundler.ts
garlicnation: Done.

src/bundler.ts
garlicnation: Done.

garlicnation:
@justinfagnani PTAL

src/bundler.ts
```diff
@@ -408,8 +441,8 @@ class Bundler {
       }
     });
 
-    // We must clone the AST from the document, since we will be modifying it.
-    const newDocument = dom5.cloneNode(analyzedRoot.parsedDocument.ast);
+    // We must parse document to a new AST since we will be modifying the AST.
+    const newDocument = dom5.parse(analyzedRoot.parsedDocument.contents);
```
justinfagnani: fyi, `parse` is removed from the latest dom5

PR #374, opened 2016-09-19
Support stripping inlined javascript

garlicnation:
Projects with complex builds might want to exclude javascript(such as polyfills that a dependency erroneously brings in).


justinfagnani:
Does this close an issue? There's very little info here on what this does and how I might know it's correct.

usergenic:
> No description provided.

Can we make sure we have one for PRs?  Also, all the tests are failing according to travis due to `TypeError: this.isStripExcludedHref is not a function`.  Is this ready for review?

garlicnation:
Holding off on this for now - it would be preferable to include this in v3 and never worry about v2.

PR #379, opened 2016-09-27
Foundational work in the sharding logic for bundles.

usergenic:
Note: this filename is `bundle-manifest` because these types/functions are working towards producing a BundleManifest object that will drive the Bundler's operations.  However, BundleManifest, is not yet implemented.

I specifically excluded the BundleManifest implementation from the branch to ensure this layer of abstraction works on its own and makes sense on its own.  Notably missing here are URL-generation for bundles and the BundleManifest interface to lookup a Bundle by a constituent file.


src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
+/**
+ * A bundle is a grouping of files which serve the need of one or more
+ * entrypoint files.
+ */
+export interface IBundle {
```
justinfagnani: Noooo.... we're not at Microsoft! :P Just `Bundle`

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+export type IBundleStrategy = (bundles: IBundle[]) => IBundle[];
+
+/**
+ * A mapping of URLs to their full set of transitive dependencies, such that
+ * `a->b, b->c, b->d, e->f` would be represented `{a:[b,c,d], e:[f]}`
+ */
+export type IFileTransDepsIndex = Map<URL, Set<URL>>;
+
+/**
+ * Defining this URL type to make it clear which strings represent URLs.
```
justinfagnani: 👍

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/**
+ * A bundle strategy function is used to transform an array of bundles.
+ */
+export type IBundleStrategy = (bundles: IBundle[]) => IBundle[];
+
+/**
+ * A mapping of URLs to their full set of transitive dependencies, such that
+ * `a->b, b->c, b->d, e->f` would be represented `{a:[b,c,d], e:[f]}`
```
justinfagnani: Good comment. Can you add a dependency shared between a and f, and also point out that a and f are the entrypoints, presumably why they're not keys in the map.

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * Defining this URL type to make it clear which strings represent URLs.
+ */
+export type URL = string;
+
+/**
+ * Standard implementation of IBundle.
+ */
+export class Bundle implements IBundle {
```
justinfagnani: A TS class also represents an interface, if that helps with naming.

src/bundle-manifest.ts
justinfagnani: I might call this `UrlString` though, since there already is a URL type which isn't a string.

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+ *   `a->b, b->c, b->d, e->c, e->f`
+ *
+ * Would produce an array of three bundles:
+ *   `[a]->[b,d], [e]->[f], [a,e]->[c]`
+ */
+export function generateBundles(depsIndex: IFileTransDepsIndex): IBundle[] {
+  const bundles: IBundle[] = [];
+  const invertedIndex = invertFileTransDepsIndex(depsIndex);
+
+  for (let [dep, entrypoints] of invertedIndex.entries()) {
```
justinfagnani: I'm not so sure about node 4's support for destructuring

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * A bundle strategy function is used to transform an array of bundles.
+ */
+export type IBundleStrategy = (bundles: IBundle[]) => IBundle[];
+
+/**
+ * A mapping of URLs to their full set of transitive dependencies, such that
+ * `a->b, b->c, b->d, e->f` would be represented `{a:[b,c,d], e:[f]}`
+ */
+export type IFileTransDepsIndex = Map<URL, Set<URL>>;
```
justinfagnani: how about `TransitiveDependencies`?

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+    return newBundles;
+  };
+}
+
+/**
+ * Returns an inverted FileTransDepsIndex where the keys are the dependencies
+ * and the values are the set of files which are dependent on them.
+ * `{a:[b,c], d:[b,e]}` becomes `{a:[a], b:[a,b,d], c:[a], d:[d], e:[d]}`.
+ */
+export function invertFileTransDepsIndex(depsIndex: IFileTransDepsIndex):
```
justinfagnani: `invertMultimap` might be a descriptive name for people familiar with collections libraries like Guava.

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+      newBundles.splice(newBundles.indexOf(sharedBundle), 1);
+      newBundles.push(mergeBundles([shellBundle, sharedBundle]));
+    }
+    return newBundles;
+  };
+}
+
+/**
+ * Returns an inverted FileTransDepsIndex where the keys are the dependencies
+ * and the values are the set of files which are dependent on them.
```
justinfagnani: ""set of files" => "set of entrypoints"

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+ * `{a:[b,c], d:[b,e]}` becomes `{a:[a], b:[a,b,d], c:[a], d:[d], e:[d]}`.
```
justinfagnani: This example has a->a in the output, but not in the input. Should it be?

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+ */
+export function invertFileTransDepsIndex(depsIndex: IFileTransDepsIndex):
+    IFileTransDepsIndex {
+  const invertedIndex = new Map<URL, Set<URL>>();
+
+  for (let [entrypoint, deps] of depsIndex.entries()) {
+    [entrypoint].concat(Array.from(deps)).forEach((dep) => {
```
justinfagnani: nit: I think `forEach` is still slower than a for/of, and since you already have a for/of it'll look more consistent

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+      if (!invertedIndex.has(dep)) {
```
justinfagnani: I like avoid `Map.has()` when you need to `get` anyway:

```typescript
let entrypoints = invertedIndex.get(dep);
if (!entrypoints) {
  entrypoints = new Set<URL>();
  invertedIndex.set(dep, entrypoints);
}
entrypoints.add(entrypoint);
```

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+ * Into the following 3 bundles, including a single bundle for all of the
+ * dependencies which are shared by at least 2 entrypoints:
+ *   `[a]->[b], [c]->[e], [a,c,f,g]->[d,h]`
+ */
+export function generateSharedDepsMergeStrategy(minEntrypoints: number):
+    IBundleStrategy {
+  return (bundles: IBundle[]): IBundle[] => {
+    const newBundles: IBundle[] = [];
+    let sharedBundle: Bundle|undefined;
+    bundles.forEach((bundle) => {
```
justinfagnani: nit: I'd use for/of

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+      if (bundle.entrypoints.size >= minEntrypoints) {
+        sharedBundle =
```
justinfagnani: maybe delay merging the shared bundles till after the loop? Let `sharedBundles` be an array...

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+            sharedBundle ? mergeBundles([sharedBundle, bundle]) : bundle;
+      } else {
+        const newBundle = new Bundle();
```
justinfagnani: do you need to create a new bundle instance? If we treat them as immutable, can we just pass them through?

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+  };
+}
+
+/**
+ * A bundle strategy function which merges all shared dependencies into a
+ * bundle for an application shell.
+ */
+export function generateShellMergeStrategy(shell: URL): IBundleStrategy {
+  return (bundles: IBundle[]): IBundle[] => {
+    const newBundles = singleSharedDepsStrategy(bundles);
```
justinfagnani: I think this should wrap `generateSharedDepsMergeStrategy` so that it takes a shell URL and a threshold. Just generate the bundles with `generateSharedDepsMergeStrategy`, then find the shared bundle and shell and merge them.

src/bundle-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+  return invertedIndex;
+}
+
+/**
+ * Given an Array of bundles, produce a single bundle with the entrypoints and
+ * files of all bundles represented.
+ */
+export function mergeBundles(bundles: IBundle[]): IBundle {
+  const newBundle = new Bundle();
+  bundles.forEach((bundle) => {
```
justinfagnani: nit: for/or

src/bundle-manifest.ts
usergenic: QQ: would you prefer maintaining an interface Bundle and then use simple objects without constructors to pass the data around or just use a Bundle class?

usergenic: I didn't realize that classes essentially worked like interfaces where type expressions were concerned.  Ditching the IBundle interface in favor of Bundle class.

src/bundle-manifest.ts
usergenic: Updated:
```
/**
 * A mapping of entrypoint to their full set of transitive dependencies,
 * such that a dependency graph `a->b, b->c, c->d, e->c, e->f` would be
 * represented `{a:[b,c,d], e:[c,f]}`.
 */
export type TransitiveDependenciesMap = Map<UrlString, Set<UrlString>>;
```

src/bundle-manifest.ts
usergenic: I went with TransitiveDependenciesMap not to be redundant, but I think because it didn't feel like a _thing_ for some reason without it.  Definitely more readable than IFileTransDepsIndex, so good call.

src/bundle-manifest.ts
usergenic: Ok.  UrlString'd it.

src/bundle-manifest.ts
usergenic: Learning that you can give compatible alternate implementations to stand in for typed values changed my perception of Typescript.  I assumed Interfaces were the only means to do that.  I like what I have learned.  Ditching IBundle.

src/bundle-manifest.ts
usergenic: Yeah this was the reason for Travis failures.  They were obscured by issues with type changes in parse5.  Thanks, guys- I was having a hard time identifying where the `[` token complaint was pointing to in the output.  *sigh*

src/bundle-manifest.ts
usergenic: I am overcoming a strong bias in favor of forEach as a result of years of Ruby Enumerable methods being based on use of .each -- I definitely see the argument for use of for/of in Javascript; Switching most/all of my forEach's to for/of loops.

src/bundle-manifest.ts
usergenic: That turns out to be cleaner looking as well as less wasteful.  Thanks!

src/bundle-manifest.ts
usergenic: You're right-- I was being overly defensive by rebuilding them.  Will just treat them as immutable.

src/bundle-manifest.ts
usergenic: Yeah.  Done'd.

src/bundle-manifest.ts
usergenic: done

src/bundle-manifest.ts
usergenic: I was trying to avoid explicit identity dependency in the initial TransitiveDependenciesMap but I don't have a good reason to keep trying to do that.  Things are cleaner if they're included.  Updating code and comments to that end.

src/bundle-manifest.ts
usergenic: Great suggestion.  Its generic now too.

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+export type UrlString = string;
+
+/**
+ * A bundle is a grouping of files which serve the need of one or more
+ * entrypoint files.
+ */
+export class Bundle {
+  public entrypoints: Set<UrlString>;
+  public files: Set<UrlString>;
+  constructor(entrypoints?: Set<UrlString>, files?: Set<UrlString>) {
```
justinfagnani: let's document whether `files` should contain the entrypoints for clarity.

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.entrypoints = entrypoints || new Set<UrlString>();
+    this.files = files || new Set<UrlString>();
+  }
+}
+
+/**
+ * Given an index of files and their dependencies, produce an array of bundles,
+ * where a bundle is defined for each set of dependencies.
```
justinfagnani: "where a bundle is defined for each set of dependencies" -> "where a bundle is defined for each set entrypoints that has unique dependencies"? I'm hunting for a clear description of this...

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+ * Would produce an array of three bundles:
+ *   `[a]->[a,b,d], [e]->[e,f], [a,e]->[c]`
+ */
+export function generateBundles(depsIndex: TransitiveDependenciesMap):
+    Bundle[] {
+  const bundles: Bundle[] = [];
+  const invertedIndex = invertMultimap(depsIndex);
+
+  for (let entry of invertedIndex.entries()) {
+    let dep: UrlString = entry[0], entrypoints: Set<UrlString> = entry[1];
```
justinfagnani: nit: I prefer one variable declaration per line.

justinfagnani: `const`

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+    const entrypointsArray = Array.from(entrypoints);
+    let bundle = bundles.find((bundle) => {
+      return bundle.entrypoints.size === entrypoints.size &&
```
justinfagnani: you can drop the `{` and `return`

justinfagnani: Actually, for readability, I'd pull this function out into something named `arrayEquals`, (or we should figure out a way to use an array as a key in a map)

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+    BundleStrategy {
+  return (bundles: Bundle[]): Bundle[] => {
+    const newBundles: Bundle[] = [];
+    const sharedBundles: Bundle[] = [];
+    for (let bundle of bundles) {
+      if (bundle.entrypoints.size >= minEntrypoints) {
+        sharedBundles.push(bundle);
+      } else {
+        newBundles.push(
+            new Bundle(new Set(bundle.entrypoints), new Set(bundle.files)));
```
justinfagnani: do you need to create a new Bundle? If `mergeBundles` doesn't mutate, but always returns a new Bundle, then it should be ok to just push `bundle`

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Inverts a map of collections such that  `{a:[c,d], b:[c,e]}` would become
+ * `{c:[a,b], d:[a], e:[b]}`.
+ */
+export function invertMultimap(multimap: Map<any, Set<any>>):
+    Map<any, Set<any>> {
+  const inverted = new Map<any, Set<any>>();
+
+  for (let entry of multimap.entries()) {
```
justinfagnani: `const` everywhere in here

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+ * Into the following 3 bundles, including a single bundle for all of the
+ * dependencies which are shared by at least 2 entrypoints:
+ *   `[a]->[a,b], [c]->[c,e], [a,c,f,g]->[d,f,g,h]`
+ */
+export function generateSharedDepsMergeStrategy(minEntrypoints: number):
+    BundleStrategy {
+  return (bundles: Bundle[]): Bundle[] => {
+    const newBundles: Bundle[] = [];
+    const sharedBundles: Bundle[] = [];
+    for (let bundle of bundles) {
```
justinfagnani: `const`

src/bundle-manifest.ts
```diff
@@ -0,0 +1,150 @@
...
+  return inverted;
+}
+
+/**
+ * Given an Array of bundles, produce a single bundle with the entrypoints and
+ * files of all bundles represented.
+ */
+export function mergeBundles(bundles: Bundle[]): Bundle {
+  const newBundle = new Bundle();
+  for (let bundle of bundles) {
```
justinfagnani: `const` throughout

src/bundle-manifest.ts

src/bundle-manifest.ts
usergenic: - Ok to drop the `{` and `return`.
- This is really a setEquals.  So I added that.

src/bundle-manifest.ts
usergenic: const'd

src/bundle-manifest.ts
usergenic: ALWAYS BE CONSTING

PR #380, opened 2016-09-28
Here's the class definition for bundle manifest and the bundle-to-url mapper function.

usergenic:
Just wanted to see if this API would work.  It's slightly different than what was proposed in the doc, so I made a separate mini-PR for consideration.

### DIFF
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -54,8 +54,8 @@ export class BundleManifest {
     for (let i = 0; i < bundles.length; ++i) {
       const bundle = bundles[i], bundleUrl = bundleUrls[i];
       this.bundles.set(bundleUrl, bundle);
-      for (const file of bundle.files) {
-        this.files.set(file, bundleUrl);
+      for (const fileUrl of bundle.files) {
+        this.files.set(fileUrl, bundleUrl);
       }
     }
   }
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -61,7 +61,7 @@ export class BundleManifest {
   }
 
   // Convenience method to return a bundle for a constituent file url.
-  getBundleForFile(url: string): Bundle|undefined {
+  getBundleForFile(url: UrlString): Bundle|undefined {
     const bundleUrl = this.files.get(url);
     if (bundleUrl) {
       return this.bundles.get(bundleUrl);
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -22,8 +22,9 @@ export type UrlString = string;
  * entrypoint files.
  */
 export class Bundle {
-  public entrypoints: Set<UrlString>;
-  public files: Set<UrlString>;
+  entrypoints: Set<UrlString>;
+  files: Set<UrlString>;
+
   constructor(entrypoints?: Set<UrlString>, files?: Set<UrlString>) {
     this.entrypoints = entrypoints || new Set<UrlString>();
     this.files = files || new Set<UrlString>();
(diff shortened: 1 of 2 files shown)


src/bundle-manifest.ts
```diff
@@ -30,6 +31,36 @@ export class Bundle {
 }
 
 /**
+ * A bundle manifest is a mapping of urls to bundles, with convenience
+ * methods to identify a bundle by a constituent file.
+ */
+export class BundleManifest {
+  public bundles: Map<UrlString, Bundle>;
```
justinfagnani: I don't think we've been doing `public`. Just elide it?

src/bundle-manifest.ts
```diff
@@ -30,6 +31,36 @@ export class Bundle {
...
+  constructor() {
+    this.bundles = new Map();
+  }
+  getMappingForFile(url: string): [UrlString, Bundle]|undefined {
```
justinfagnani: `url: UrlString`?

justinfagnani: You could invert the bundles map for quick lookup of file->bundle

src/bundle-manifest.ts
```diff
@@ -61,6 +92,22 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
...
 /**
+ * Given a collection of bundles and a BundleUrlMapper to generate urls for
+ * them, produce a BundleManifest.
+ */
+export function generateBundleManifest(
+    bundles: Bundle[], urlGenerator: BundleUrlMapper): BundleManifest {
+  const urls = urlGenerator(bundles);
+  const manifest = new BundleManifest();
+
+  for (let i = 0; i < bundles.length; ++i) {
```
justinfagnani: `const`

src/bundle-manifest.ts
```diff
@@ -148,3 +195,19 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
   return newBundle;
 }
+
+/**
+ * A simple function for generating shared bundle names based on a counter.
+ */
+export function sharedBundleUrlMapper(bundles: Bundle[]): UrlString[] {
+  let counter = 0;
+  const urls: UrlString[] = [];
+  for (let bundle of bundles) {
```
justinfagnani: `const`

src/bundle-manifest.ts
```diff
@@ -61,6 +92,22 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
 }
 
 /**
+ * Given a collection of bundles and a BundleUrlMapper to generate urls for
+ * them, produce a BundleManifest.
+ */
+export function generateBundleManifest(
```
justinfagnani: might be a good static method on `BundleManifest`, or even a constructor.

src/bundle-manifest.ts
usergenic: Pulling out of this and Bundle class then.  Yay implicitness!

src/bundle-manifest.ts
usergenic: Dunno why I skipped that.  Smaller and cleaner code.

src/bundle-manifest.ts
usergenic: I also have old bias against constructor logic, but I think it makes sense to use constructor here.  Its cleaner.

src/bundle-manifest.ts
usergenic: I did not realize you could const this.  I have lots of `for (let x of xs)` and `let x` inside of for loops that I can const.  CONST ALL THE THINGS!  Done.

usergenic:
I must insist upon LGTM!

src/bundle-manifest.ts
```diff
@@ -21,15 +22,54 @@ export type UrlString = string;
...
+    const bundleUrls = urlMapper(bundles);
+
+    this.bundles = new Map();
+    this.files = new Map();
+
+    for (let i = 0; i < bundles.length; ++i) {
+      const bundle = bundles[i], bundleUrl = bundleUrls[i];
+      this.bundles.set(bundleUrl, bundle);
+      for (const fileUrl of bundle.files) {
+        this.files.set(fileUrl, bundleUrl);
```
justinfagnani: maybe assert `!this.files.has(fileUrl)`

src/bundle-manifest.ts
```diff
@@ -21,15 +22,54 @@ export type UrlString = string;
...
 
 /**
+ * A bundle manifest is a mapping of urls to bundles.
+ */
+export class BundleManifest {
+  // Map of bundle url to bundle.
+  bundles: Map<UrlString, Bundle>;
+
+  // Map of file url to bundle url.
+  files: Map<UrlString, UrlString>;
```
justinfagnani: The name `files` with the same type for key and value could be a little vague. A name like `bundleUrlForFile` might read better at the use site: `bundleManifest.bundleUrlForFile.get(url)`.

The other option is to make this private and expose a method: `bundleManifest.getBundleUrlForFile(url)`

src/bundle-manifest.ts
usergenic: Good point.  In case bundle strategy or other part of bundle definition pipeline is broken.

src/bundle-manifest.ts

PR #386, opened 2016-10-05
New url mapper

garlicnation:


src/bundle-manifest.ts
```diff
@@ -219,15 +216,85 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+    const reduced = new Set<T>();
+    for (let entry of previous) {
+      if (current.has(entry)) {
+        reduced.add(entry);
+      }
     }
+    return reduced;
+  });
+}
+
```
usergenic: Comment

src/bundle-manifest.ts
```diff
@@ -219,15 +216,85 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
-      urls.push(Array.from(bundle.entrypoints)[0]);
+      urlMap.set(Array.from(bundle.entrypoints)[0], bundle);
     } else {
-      urls.push(`shared_bundle_${++counter}.html`);
+      urlMap.set(`shared_bundle_${++counter}.html`, bundle);
+    }
+  }
+  return urlMap;
+}
+
```
usergenic: Comment

src/bundle-manifest.ts
```diff
@@ -219,15 +216,85 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+  return list;
+}
+
+/**
+ * Attempts to generate shared bundle names by identifying unique endpoints,
+ * falling back on `sharedBundleUrlMapper` when unable to form a direct
+ * correspondence.
+ *
+ * example:
+ * const bundle1 = {endpoints: ['A']};
```
garlicnation: s/endpoint/entrypoint/

src/bundle-manifest.ts
```diff
@@ -219,15 +216,85 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+function getEntrypointSets(bundles: Bundle[]): Set<string>[] {
+  const list: Set<string>[] = [];
+  for (let bundle of bundles) {
+    list.push(bundle.entrypoints);
+  }
+  return list;
+}
+
+/**
+ * Attempts to generate shared bundle names by identifying unique endpoints,
```
usergenic: You are using the word `endpoints` here instead of `entrypoints`.  Is this a mistake or does this mean something different?

src/bundle-manifest.ts
```diff
@@ -49,14 +49,11 @@ export class BundleManifest {
...
-    const bundleUrls = urlMapper(bundles);
-
-    this.bundles = new Map();
+    this.bundles = urlMapper(bundles);
     this.bundleUrlForFile = new Map();
 
-    for (let i = 0; i < bundles.length; ++i) {
-      const bundle = bundles[i], bundleUrl = bundleUrls[i];
-      this.bundles.set(bundleUrl, bundle);
+    for (let bundleMapEntry of this.bundles) {
```
usergenic: `for (const bundleMapEntry of this.bundles) {`

src/bundle-manifest.ts
```diff
@@ -219,15 +216,85 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+ * becomes:
+ * ['A.html', 'B.html', 'C.html']
+ */
+export function uniqueEndpointUrlMapper(bundles: Bundle[]):
+    Map<UrlString, Bundle> {
+  const bundleMap = new Map<UrlString, Bundle>();
+  // Avoid mutating passed array;
+  const remainingBundles = Array.from(bundles);
+  // Variable to track if an iteration of the loop resulted in a new bundle
+  // ssignment.
```
usergenic: assignment

src/bundle-manifest.ts

garlicnation: Done

src/bundle-manifest.ts
garlicnation: Done

src/bundle-manifest.ts
garlicnation: Done

src/bundle-manifest.ts
garlicnation: Done

src/bundle-manifest.ts
garlicnation: Renamed

garlicnation:
@usergenic This only computes names. Bundle contents aren't moved around at all by this change. The intersections happen here because it's a convenient heuristic which entrypoint should represent which bundle.

garlicnation:
@usergenic PTAL

usergenic:
@garlicnation you're right-- I was mistaken about the `uniqueEntrypointUrlMapper`.

I was thinking about it though, and we have a general issue in the bundle generation system with respect to entrypoints having dependencies on other entrypoints.  If we added the behavior such that the following was true...
```
    test('polymer core generate case', () => {
      const depsIndex = new Map<string, Set<string>>();
      depsIndex.set('P', new Set(['P', '1', 'PMini', '2', 'PMicro', '3']));
      depsIndex.set('PMini', new Set(['PMini', '2', 'PMicro', '3']));
      depsIndex.set('PMicro', new Set(['PMicro', '3']));

      const bundles = generateBundles(depsIndex).map(serializeBundle).sort();
      console.log(bundles);
      assert.equal(bundles.length, 3);
      assert.equal(bundles[0], '[P]->[1,P]');
      assert.equal(bundles[1], '[PMicro]->[2,PMicro]');
      assert.equal(bundles[2], '[PMini]->[3,PMini]');
    });
```
...then you wouldn't need a mapper to work so hard to name things and *also* the strategies would correctly honor the hard-boundary of entrypoints.

Right now if you run a `sharedDepsMergeStrategy(2)` on `[a,b,c]->[a], [b,c]->[b], [c]->[c]` you wind up with `[a,b,c]->[a,b], [c]->[c]` which is bad.  Entrypoints should have a kind of primacy that forbids this and I think the correct place to address it is in the generateBundles function.

PR #387, opened 2016-10-07
Updated generateBundles

usergenic:
I updated generateBundles to honor the exclusivity of entrypoints, so that an entrypoint is not bundled with another by default.

There are tests to demonstrate.

### DIFF
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -100,11 +100,43 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
     Bundle[] {
   const bundles: Bundle[] = [];
 
-  // TODO(usergenic): Assert a valid transitive dependencies map; i.e.
-  // entrypoints should include themselves as dependencies and entrypoints
-  // should *probably* not include other entrypoints as dependencies.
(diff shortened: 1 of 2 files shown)


src/bundle-manifest.ts
```diff
@@ -100,11 +100,12 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
     Bundle[] {
   const bundles: Bundle[] = [];
 
-  // TODO(usergenic): Assert a valid transitive dependencies map; i.e.
-  // entrypoints should include themselves as dependencies and entrypoints
-  // should *probably* not include other entrypoints as dependencies.
+  const prunedDepsIndex: TransitiveDependenciesMap =
+      pruneTransitiveDependenciesMap(depsIndex);
```
garlicnation: I wonder if this should be optional. In some cases it makes sense to merge entrypoints together. I'm worried that if the depsIndex passed to the bundle heuristics is always pruned of entrypoints and their transitive dependencies, it will be much more challenging to perform these optimizations

usergenic: This is merging to your newMapper branch fyi

PR #388, opened 2016-10-10
Use manifest logic for all bundling

garlicnation:
Single file and multiple file bundling use the same module-based paths. The sharding test passes and all the single file tests pass.


src/deps-index.ts
```diff
@@ -69,6 +69,7 @@ export async function buildDepsIndex(entrypoints: string[], analyzer: Analyzer):
         }
         const dependencyEntry =
             await _getTransitiveDependencies(entrypoint, analyzer);
+        debugger;
```
usergenic: remove debugger

src/bundler.ts
```diff
@@ -408,23 +426,54 @@ class Bundler {
...
+    }
   }
 
   private async _bundleDocument(
       url: string,
-      excludes: string[],
-      stripExcludes: string[]): Promise<ASTNode> {
+      bundle: Bundle,
+      bundleManifest: BundleManifest,
+      addedImports?: Set<string>): Promise<ASTNode> {
```
usergenic: This name leads me to think the imports provided in the set are additive (i.e. not represented in the document already)... is that the intent?  If so, [line 465](https://github.com/Polymer/vulcanize/pull/388/files#diff-841ba21ca2afd1b3c22749d1d7190395R465) where we hand `bundle.files` in as this argument was confusing to me.  I guess its called added because it might include some that aren't already in the doc.  But also, it defines whether inlining of an import could even happen, right?  So maybe its more like an `override` than an `added`?  Anyways, maybe add a comment?  Comment could just be above [line 509](https://github.com/Polymer/vulcanize/pull/388/files#diff-841ba21ca2afd1b3c22749d1d7190395R509) `if (addedImports) {`

src/bundle-manifest.ts
```diff
@@ -219,15 +220,123 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+  });
+}
+
+/**
+ * Returns a list of the sets of entrypoints, for use in performing
+ * set intersection and subtraction on the sets.
+ */
+function getEntrypointSets(bundles: Bundle[]): Set<string>[] {
+  const list: Set<string>[] = [];
+  for (let bundle of bundles) {
```
garlicnation: Done

src/bundler.ts
```diff
@@ -408,23 +426,54 @@ class Bundler {
...
-    collection.set(entrypoints[0], doc);
-    return collection;
+  async bundle(
+      entrypoints: string[],
+      strategy?: BundleStrategy,
+      mapper?: BundleUrlMapper): Promise<DocumentCollection> {
+    const bundledDocuments = new Map<string, ASTNode>();
+    if (entrypoints.length === 1) {
+      const url = entrypoints[0];
+      debugger;
```

src/bundler.ts
```diff
@@ -408,23 +426,54 @@ class Bundler {
...
+      const depsIndex = await buildDepsIndex(entrypoints, this.analyzer);
+      debugger;
```

src/bundler.ts
```diff
@@ -408,23 +426,54 @@ class Bundler {
...
+      const doc = await this._bundleDocument(url, bundles[0], manifest);
+      bundledDocuments.set(url, doc);
+      return bundledDocuments;
+    } else {
+      const bundles = new Map<string, ASTNode>();
+      if (!strategy) {
+        console.assert(strategy, 'strategy must be provided!');
+        throw new Error('strategy must be provided');
+      }
+      if (!mapper) {
```

src/deps-index.ts
garlicnation: Done

src/bundler.ts
garlicnation: Added a comment explaining what's going on.

src/ast-utils.ts
```diff
@@ -50,3 +52,35 @@ export function siblingsAfter(node: ASTNode): ASTNode[] {
...
+    const content = treeAdapters.default.getTemplateContent(template);
+    for (const match of dom5.queryAll(content, predicate)) {
+      results.push(match);
+    }
+  }
+  return results;
+}
+
+/**
+ * Find content inside <template> tags.
```
justinfagnani: does this have to be called on a template node? Or does it just find everything, even across template content?

src/ast-utils.ts
```diff
@@ -50,3 +52,35 @@ export function siblingsAfter(node: ASTNode): ASTNode[] {
...
+ */
+export function querySelectorAllTemplates(
+    node: ASTNode, predicate: dom5.Predicate): ASTNode[] {
+  const results: ASTNode[] = [];
+  const templates = dom5.queryAll(node, matchers.template);
+  for (const template of templates) {
+    const content = treeAdapters.default.getTemplateContent(template);
```
justinfagnani: what's the benefit of this API over `template.content`?

src/ast-utils.ts
```diff
@@ -50,3 +52,35 @@ export function siblingsAfter(node: ASTNode): ASTNode[] {
...
+    for (const match of dom5.queryAll(content, predicate)) {
+      results.push(match);
+    }
```
justinfagnani: should this loop recurse into sub-templated? Please doc whether or not this function recurses.

src/bundle-manifest.ts
```diff
@@ -49,14 +52,12 @@ export class BundleManifest {
...
-
-    this.bundles = new Map();
+    this.bundles = urlMapper(Array.from(bundles));
     this.bundleUrlForFile = new Map();
 
-    for (let i = 0; i < bundles.length; ++i) {
-      const bundle = bundles[i], bundleUrl = bundleUrls[i];
-      this.bundles.set(bundleUrl, bundle);
+    for (const bundleMapEntry of this.bundles) {
+      const bundleUrl = bundleMapEntry[0], bundle = bundleMapEntry[1];
```
justinfagnani: style nit: one declaration per line

src/bundle-manifest.ts
```diff
@@ -49,14 +52,12 @@ export class BundleManifest {
...
+      bundle.url = bundleUrl;
```
justinfagnani: Seems odd to mutate the inputs to the constructor. Can bundles be created with their URL, or keep this in a map?

src/bundle-manifest.ts
```diff
@@ -219,15 +220,123 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+  // assignment.
+  let assignedBundle = true;
+
+  /**
+   * Attempt to find a name, and bail once a search
+   * round goes without a candidate.
+   *
+   * Given N sets of entrypoints(strings), names are selected as follows.
+   *
+   * Name_i = Any member of
```
justinfagnani: I'm not sure what this means. Example?

src/bundle-manifest.ts
```diff
@@ -219,15 +220,123 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+  return list;
 }
+
+/**
+ * Attempts to generate shared bundle names by identifying unique Entrypoints,
+ * falling back on `sharedBundleUrlMapper` when unable to form a direct
+ * correspondence.
+ *
+ * example:
+ * const bundle1 = {Entrypoints: ['A']};
```
justinfagnani: Is this for the polymer core build use case? I think we determined that we could name bundles by looking for entrypoints in the _dependencies_ of a bundle.

Polymer core should cause a bundle list of something like:

(std): [std, a, b, c]
(std, mini): [mini, d, e, f]
(std, mini, micro): [micro, g, h, i]

As long are there aren't any dependency cycles between entrypoints, there will be only one entrypoint in the dependencies list, and you can simply name the bundle after that.

src/ast-utils.ts
usergenic: ASTNode doesn't expose `content` as a property so tslint throws fit (correctly) if we directly reference.

src/bundle-manifest.ts
```diff
@@ -219,15 +220,123 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
+}
+
+/**
+ * Compute the intersection of all of the passed `sets`.
+ * See: https://en.wikipedia.org/wiki/Intersection_(set_theory)
+ */
+function setIntersection<T>(sets: Set<T>[]): Set<T> {
+  return sets.reduce((previous, current) => {
+    const reduced = new Set<T>();
+    for (let entry of previous) {
```

src/bundler.ts
```diff
@@ -408,23 +426,54 @@ class Bundler {
...
+        bundles[0].files.delete(exclude);
+      }
+      const manifest =
+          new BundleManifest(bundles, () => new Map([[url, bundles[0]]]));
+      const doc = await this._bundleDocument(url, bundles[0], manifest);
+      bundledDocuments.set(url, doc);
+      return bundledDocuments;
+    } else {
+      const bundles = new Map<string, ASTNode>();
+      if (!strategy) {
```
garlicnation: Added a default strategy.

src/ast-utils.ts
garlicnation: What @usergenic said. I can get a typings PR submitted, but using this API seems like the path of least resistance for now.

src/bundle-manifest.ts
garlicnation: Replaced with entrypoint membership algorithm.

src/bundle-manifest.ts
garlicnation: Replace this function with this suggestion.

src/bundle-manifest.ts
garlicnation: I created a new type w/ a bundle and a URL. It's very convenient to keep them together at various points.

src/ast-utils.ts
garlicnation: Documented and added an option.

src/bundle-manifest.ts
garlicnation: Fixed.

PR #389, opened 2016-10-11
Refactor

garlicnation:
Clean up some logic around loading resources. CSS/js resources that fail to load are now automatically excluded.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -78,6 +78,11 @@ export interface Options {
   stripExcludes?: string[];
 }
 
+class BundledDocument {
+  hiddenDiv: ASTNode;
+  document: ASTNode;
+}
+
 class Bundler {
   basePath?: string;
   addedImports: string[];
@@ -185,21 +190,18 @@ class Bundler {
   /**
    * Inline external scripts <script src="*">
    */
-  inlineScript(
-      docUrl: string,
-      externalScript: ASTNode,
-      importMap: Map<string, Import|null>): ASTNode|undefined {
+  async inlineScript(docUrl: string, externalScript: ASTNode):
+      Promise<ASTNode|undefined> {
     const rawUrl: string = dom5.getAttribute(externalScript, 'src')!;
     const resolvedUrl = urlLib.resolve(docUrl, rawUrl);
-    const script = importMap.get(resolvedUrl);
+    const script = await this.analyzer.analyze(resolvedUrl);
 
-    if (!script || !script.document) {
+    if (!script) {
       return;
     }
 
     // Second argument 'true' tells encodeString to escape <script> tags.
-    const scriptContent =
-        encodeString(script.document.parsedDocument.contents, true);
+    const scriptContent = encodeString(script.parsedDocument.contents, true);
     dom5.removeAttribute(externalScript, 'src');
     dom5.setTextContent(externalScript, scriptContent);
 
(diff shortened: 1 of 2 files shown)


src/bundler.ts
```diff
@@ -483,25 +510,7 @@ class Bundler {
     // feature as a value indicating the inlining of the Import has not yet
     // occurred or a value of null indicating that <link> tags referencing it
     // should be removed from the document.
```
usergenic: This comment should be changed or removed now that variable changed.

src/bundler.ts
```diff
@@ -293,24 +321,23 @@ class Bundler {
...
     }
 
-    // If we've just inlined it or otherwise have seen it before, we can remove
-    // the <link> tag.
-    if (importMap.get(resolvedUrl) === null) {
+    if (reachedImports.has(resolvedUrl)) {
       dom5.remove(htmlImport);
     }
 
     // If we've never seen this import before, lets put it on the map as null so
```
usergenic: Change to
> If we've never seen this import before, lets add it to the set so we will deduplicate if we encounter it again.

src/bundler.ts
```diff
@@ -293,24 +321,23 @@ class Bundler {
...
 
       astUtils.insertAllBefore(importParent, htmlImport, importDoc.childNodes!);
     }
 
-    // If we've just inlined it or otherwise have seen it before, we can remove
-    // the <link> tag.
-    if (importMap.get(resolvedUrl) === null) {
+    if (reachedImports.has(resolvedUrl)) {
       dom5.remove(htmlImport);
     }
```
usergenic: Should the following `if (!reachedImports.has(resolvedUrl)` maybe be converted to an else clause of this if statement?

src/bundler.ts
```diff
@@ -542,10 +551,8 @@ class Bundler {
             newUrl, 'html-import', importDoc, undefined, undefined, newNode, [
             ]);
```
usergenic: I don't think this `const im = new Import` thing needs to exist anymore...?

src/bundler.ts
```diff
@@ -568,32 +575,35 @@ class Bundler {
       }
     });
 
+    const reachedImports = new Set<UrlString>([url]);
+
     // Inline all HTML Imports.  (The inlineHtmlImport method will discern how
     // to handle them based on the state of the importMap.)
```
usergenic: Fix comment; importMap no longer a thing.

usergenic:
I meant to submit as "Request Changes" but I don't see an undo for that on Github.  Most of the issues are minor so keeping this as LGTM is fine-- just have a look at the comments I made.

src/bundler.ts
garlicnation: Done.

src/bundler.ts
garlicnation: Done.

src/bundler.ts
garlicnation: Done.

src/bundler.ts

src/bundler.ts
garlicnation: Done.

src/bundler.ts
```diff
@@ -243,33 +254,50 @@ class Bundler {
...
-    //   dom5.setAttribute(htmlImport, 'href', bundleUrl);
-    // }
+    if (!bundleUrl) {
+      if (reachedImports.has(resolvedUrl)) {
+        dom5.remove(htmlImport);
+        return;
+      } else {
+        reachedImports.add(resolvedUrl);
+      }
+      // dom5.remove(htmlImport);
```
usergenic: remove commented-out code.

PR #391, opened 2016-10-18
automatic lazy bundle support

garlicnation:
Add the ability for vulcanize to determine bundles based on lazy imports detected through <link rel="lazy-import" tags.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -467,24 +467,6 @@ class Bundler {
       mapper?: BundleUrlMapper): Promise<DocumentCollection> {
     const bundledDocuments: DocumentCollection =
         new Map<string, BundledDocument>();
-    if (entrypoints.length === 1) {
-      const url = entrypoints[0];
-      const depsIndex = await buildDepsIndex(entrypoints, this.analyzer);
-      const bundles = generateBundles(depsIndex.entrypointToDeps);
-      for (const exclude of this.excludes) {
-        bundles[0].files.delete(exclude);
-      }
-      const manifest =
-          new BundleManifest(bundles, () => new Map([[url, bundles[0]]]));
-      const bundle = {
-        url: url,
-        bundle: bundles[0],
-      };
-      const doc = await this._bundleDocument(bundle, manifest);
-      bundledDocuments.set(
-          url, {ast: doc, files: Array.from(bundles[0].files)});
-      return bundledDocuments;
-    } else {
       const bundles = new Map<string, ASTNode>();
       if (!strategy) {
         strategy = generateSharedDepsMergeStrategy(2);
@@ -506,7 +488,6 @@ class Bundler {
             {ast: bundledAst, files: Array.from(bundle.bundle.files)});
       }
       return bundledDocuments;
-    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/bundler.ts
```diff
@@ -262,15 +260,15 @@ class Bundler {
...
     const resolvedUrl: string = urlLib.resolve(docUrl, rawUrl);
-    const bundleUrl = manifest.bundleUrlForFile.get(resolvedUrl);
+    const bundleUrl = this.resolveBundleUrl(resolvedUrl, bundle, manifest);
     if (!bundleUrl) {
       if (reachedImports.has(resolvedUrl)) {
         dom5.remove(htmlImport);
         return;
       } else {
         reachedImports.add(resolvedUrl);
+        return;
```
justinfagnani: I kind of like the return in the original position, and remove the one above

src/bundler.ts
```diff
@@ -467,46 +465,26 @@ class Bundler {
...
+    }
+    const index = await buildDepsIndex(entrypoints, this.analyzer);
+    const basicBundles = generateBundles(index.entrypointToDeps);
+    const bundlesAfterStrategy = strategy(basicBundles);
+    const manifest = new BundleManifest(bundlesAfterStrategy, mapper);
+    for (const bundleEntry of manifest.bundles) {
+      const bundleUrl = bundleEntry[0];
+      const bundle = {url: bundleUrl, bundle: bundleEntry[1]};
+      const bundledAst =
+          await this._bundleDocument(bundle, manifest, bundle.bundle.files);
```
justinfagnani: `bundle.bundle`? That's unfortunate. Are there more specific names that would help give meaning to that?

src/bundler.ts
```diff
@@ -126,26 +126,24 @@ class Bundler {
...
       url: string,
       bundle: AssignedBundle,
-      manifest: BundleManifest): boolean|string {
+      manifest: BundleManifest): string|false {
     const targetBundle = manifest.getBundleForFile(url);
     if (!targetBundle || !targetBundle.url) {
       return false;
     }
+    for (const exclude of this._excludes) {
+      if (url.search(exclude) > -1) {
```
justinfagnani: `String.prototype.search` takes a RegExp, and will convert a string into a RegExp via `new RexExp`. That could lead to some funky behavior with some strings.

garlicnation:
It works because most of the support was added in the analyzer, and lazy-imports is simply treated as a special case where entrypoints can be inferred.

stramel:
So what do we need to do to get this in?

justinfagnani:
@stramel now that the v3 branch is merged into master, we may want to take a fresh look. I'm going to retarget this PR to master so we can preserve it and delete the v3 branch. Then @usergenic will take a look, maybe this week.

stramel:
Great! Thanks for the update @justinfagnani!

usergenic:
I am looking at this PR and after resolving conflicts locally, it doesn't appear to have any actual changes in it.  The `deps-index.ts` code already supports the `lazy-html-import` type and the bundle url resolution w/ excludes is already supported in another location.  There might be something subtle I'm missing, but I'm going to take a different tack and create a separate PR with tests to demonstrate handling of `<link rel="lazy-import"...>` since that's what's missing.

usergenic:
@stramel and @justinfagnani https://github.com/Polymer/polymer-bundler/pull/487 supercedes this now

Issue #368, opened 2016-09-06
resolveUrl returns wrong path when serving with build/bundled


johnlim:
Hi, 
I opened an issue initially with [polymer-starter-kit](https://github.com/PolymerElements/polymer-starter-kit/issues/913) but its probably a vulcanize issue. I've copied/pasted the issue over here for convenience.

### Description

Calling resolveUrl in a nested element returns a different url when serving with `polymer serve` and `polymer serve build/bundled`
### Expected outcome

When calling `polymer serve build/bundled`, resolveUrl should return url with base url
### Actual outcome

When calling `polymer serve`, resolveUrl returns `http://localhost:8080/src/custom-element.html` but `polymer serve build/bundled` simply returns `custom-element.html`
### Live Demo

https://github.com/johnlim/polymer-starter-kit/tree/resolveUrl
### Steps to reproduce

1) clone the repo and switch to the branch resolveUrl
2) run bower install if necessary
3) run `polymer build` 
4) run `polyserve` and observe console output which prints out the resolved url as `http://localhost:8080/src/custom-element.html`
5) Clear application cache in browser
6) run `polyserve build/bundled` and observe console output which prints out the resolved url `custom-element.html`
### Browsers Affected

Only tried in Chrome

- [X ] Chrome

usergenic:
Assigning to myself to verify whether this is still an issue.  There've been a lot of releases since this was filed.

usergenic:
Just ran `polymer build --bundle` with polymer-cli 1.5.5 and the console output is identical for bundled vs unbundled case.  Closing issue -- Please reopen if this is still a problem.

Thanks

PR #397, opened 2016-11-09
assetpath in same directory fix

josh-bradley:
Fix to Issue #368. This issue occurs when the entry point file imports a component from the same directory, the assetpath attribute for the component gets set to '/', this in turn results in the wrong value being returned from the resolveUrl function.

The fix makes it so the assetpath is set to an empty string. Could also be './' I wasn't sure which would be best.


usergenic:
I'm going to modify PR to pull this into remote branch for a couple additional fixes before master merge.  Thanks, @josh-bradley

PR #400, opened 2016-11-28
fix file resolving to work with root-level files

FredKSchott:
This is a small fix related to how URLs are resolved. This was the visible bug that caused all the trouble in https://github.com/Polymer/polymer-build/pull/83.

/cc @justinfagnani @usergenic @rictic

### DIFF
--- a/lib/pathresolver.js
+++ b/lib/pathresolver.js
@@ -86,7 +86,14 @@ PathResolver.prototype = {
     var parsedFrom = url.parse(mainDocUrl);
     var parsedTo = url.parse(absUrl);
     if (parsedFrom.protocol === parsedTo.protocol && parsedFrom.host === parsedTo.host) {
-      var pathname = pathPosix.relative(pathPosix.dirname(parsedFrom.pathname), parsedTo.pathname);
+      var parsedFromDir = pathPosix.dirname(parsedFrom.pathname);
+      // top-level file paths like `index.html` will return the dirname `.`.
+      // To calculate the relative path below, we should resolve against the
+      // root (`/`) instead.
+      if (parsedFromDir === '.') {
+        parsedFromDir = path.sep;
+      }
+      var pathname = pathPosix.relative(parsedFromDir, parsedTo.pathname);
       return url.format({
         pathname: pathname,
         search: parsedTo.search,
--- a/test/html/absolute-imports.html
+++ b/test/html/absolute-imports.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<!--
+    @license
+    Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+    This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+    The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+    The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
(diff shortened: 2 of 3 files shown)


lib/pathresolver.js
```diff
@@ -86,7 +86,14 @@ PathResolver.prototype = {
     var parsedFrom = url.parse(mainDocUrl);
     var parsedTo = url.parse(absUrl);
     if (parsedFrom.protocol === parsedTo.protocol && parsedFrom.host === parsedTo.host) {
-      var pathname = pathPosix.relative(pathPosix.dirname(parsedFrom.pathname), parsedTo.pathname);
+      var parsedFromDir = pathPosix.dirname(parsedFrom.pathname);
+      // top-level file paths like `index.html` will return the dirname `.`.
+      // To calculate the relative path below, we should resolve against the
+      // root (`/`) instead.
+      if (parsedFromDir === '.') {
```
usergenic: FWIW, the new v3 branch bundler code works around this with a custom URL relative calculation function that incorporates the logic for isAbsoluteUrl checks etc., so I don't want to get too in depth on this fix-- just trying to understand if there is a case where we expect top-level path not to include the leading '/' -- The tests don't seem to have a case for this, so I just wasn't sure how this branch gets exercised...?  Wouldn't this only happen if we were testing relative imports, not absolute imports?  This PR added '/html/absolute-imports.html'.

I feel like I am missing something... HALP!

lib/pathresolver.js
```diff
@@ -86,7 +86,14 @@ PathResolver.prototype = {
     var parsedFrom = url.parse(mainDocUrl);
     var parsedTo = url.parse(absUrl);
     if (parsedFrom.protocol === parsedTo.protocol && parsedFrom.host === parsedTo.host) {
-      var pathname = pathPosix.relative(pathPosix.dirname(parsedFrom.pathname), parsedTo.pathname);
+      var parsedFromDir = pathPosix.dirname(parsedFrom.pathname);
+      // top-level file paths like `index.html` will return the dirname `.`.
```
justinfagnani: I find this comment a little hard to grok. What's a top-level path, and where do we get it?

lib/pathresolver.js
usergenic: For reference, the new relativeUrl function in v3 for reference.  https://github.com/Polymer/vulcanize/blob/v3/src/url-utils.ts#L48
https://github.com/Polymer/vulcanize/blob/v3/src/test/url-utils_test.ts#L67

None of tests address a case where pathname of the "from" URL isn't absolute, but I just tried calling the new relativeUrl with args 'index.html', 'some-other-thing.html' and it did the right thing, so I think we'll not see a regression in build when we switch to v3.

lib/pathresolver.js
```diff
@@ -86,7 +86,14 @@ PathResolver.prototype = {
...
+      // To calculate the relative path below, we should resolve against the
+      // root (`/`) instead.
+      if (parsedFromDir === '.') {
+        parsedFromDir = path.sep;
```
FredKSchott: right, good catch

FredKSchott:
No longer needed for polymer-build thanks to https://github.com/Polymer/polymer-build/pull/83. 

This may still be a problem for other projects, but given the in-progress v3 branch where this is already fixed, I don't think it's worth potentially breaking things for others now.

PR #401, opened 2016-12-07
Add --out-request-list option

danbeam:
- [X] CHANGELOG.md has been updated

Useful for calculating inputs for build tools

### DIFF
--- a/bin/vulcanize
+++ b/bin/vulcanize
@@ -82,7 +82,8 @@ var args = nopt(
     'no-implicit-strip': Boolean,
     'inline-scripts': Boolean,
     'inline-css': Boolean,
-    'out-html': String
+    'out-html': String,
+    'out-request-list': String
   },
   {
     'h': ['--help'],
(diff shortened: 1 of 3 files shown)


justinfagnani:
We need to make sure this is added to the v3 branch as well.

justinfagnani:
Hmm, looking at what this does, I'm not sure it's really a Vulcanize feature rather than an analyzer feature. Vulcanize just happens to run analysis and know what files are referenced.

usergenic:
@justinfagnani I _think_ this is a bundler feature specifically because it relates to the bundled files instead of just the referenced files.

usergenic:
I.e. it has to honor the flags for the bundler not just analyzer.

web-padawan:
+1 to this feature, I think this might be useful for [webcomponent-bundle-analyzer](https://github.com/samccone/webcomponent-bundle-analyzer)

justinfagnani:
That's the thing, webcomponent-bundle-analyzer just uses the analyzer directly.

web-padawan:
Hmm, you're right. Just noticed that webcomponent-bundle-analyzer doesn't take into account sharding logic at all.

danbeam:
fwiw: I'm thinking --list-urls or --list-requests might be more accurate.  thoughts?

bin/vulcanize
```diff
@@ -126,15 +128,25 @@ args.implicitStrip = !args['no-implicit-strip'];
 args.inlineScripts = args['inline-scripts'];
 args.inlineCss = args['inline-css'];
 
-(new vulcan(args)).process(target, function(err, content) {
+var v = new vulcan(args);
```
justinfagnani: let's avoid single-letter variable names, just use `vulcanize`

justinfagnani:
I think maybe `--out-request-list`, since this needs to take a filename, like `--out-html`.

danbeam:
should I be using https://github.com/Polymer/polymer-bundler/tree/1.x as the upstream for this PR now?

justinfagnani:
Yeah, that's the Vulcanize branch

justinfagnani:
@danbeam just in case you didn't know, you don't have to create a new PR to change the base branch, you can just edit the PR and choose a new base.

bin/vulcanize
```diff
@@ -126,15 +128,22 @@ args.implicitStrip = !args['no-implicit-strip'];
...
+var vulcanize = new vulcan(args);
+vulcanize.process(target, function(err, content) {
   if (err) {
     process.stderr.write(require('util').inspect(err));
     process.exit(1);
   }
+
+  if (args['out-request-list']) {
+    var urls = Object.keys(vulcanize.loader.requests).filter(function(url) {
+      return !vulcanize.isExcludedHref(url);
```
danbeam: @freshp86 and I felt this was saner overall

danbeam:
p.s. if you need me to squash the commit history I can do that :smile:

danbeam:
Squashed the history.  Is there anybody out there?

azakus:
@usergenic @FredKSchott

Issue #357, opened 2016-08-05
Style includes in base document need to be reordered


rictic:
`index.html`

``` html
<head>
<link rel="import" href="bower_components/polymer/polymer.html">
<link rel="import" href="dep.html">
<style is="custom-style" include="foo"></style>
</head>
<body>
<div>
  hello
</div>
</body>
```

`dep.html`

``` html
<dom-module id='foo'>
  <template>
    <style>
      div {
        color: red;
      }
    </style>
  </template>
</dom-module>
```

When unvulcanized the `<style include>` in index.html comes after the `<link import>`, and so the include is able to resolve and the body text is red.

Vulcanization however produces this:

`vulcanized.html`

``` html
<html><head><meta charset="UTF-8">
<style is="custom-style" include="foo"></style>
</head>
<body><div hidden="" by-vulcanize="">

<!-- ...snip contents of polymer.html... -->

<dom-module id="foo" assetpath="/">
  <template>
    <style>
      div {
        color: red;
      }
    </style>
  </template>
</dom-module>
</div>
<div>
  hello
</div>
</body></html>
```

Now because the style include comes _before_ the style module it's including, the body text is unstyled.

I believe the correct thing to do here would be to move all styles, or at least all style includes, from the main document so that they're located after the end of the inlined documents.

maxiplay:
+1 its break my style on firefox

Bubbit:
It seems to be related to this change: https://github.com/Polymer/vulcanize/commit/4511f01bd23e27edd407c443ec2630ef91a6ee25#diff-2fd25663c1d3dca2153ce60dac0f7245R179

usergenic:
Vulcanize 1.15.1 released.  Should fix this.

Issue #375, opened 2016-09-23
Polymer source custom-styles get overrided by dependancy custom-styles after vulcanize


ThilinaTCH:
Hi All,

We have a custom-style which applies some root level styles. We are keeping this in a separate html. Following is a sample of it,
`<link rel="import" href="../bower_components/polymer/polymer.html">

<style is="custom-style">

```
:root {
    --primary-text-color: var(--light-theme-text-color);
    }
```

</style>`

Also there is a dependency with paper-styles which has same kind of custom style which set some of same styles we are using (for ex --primary-text-color)

After vulcanization happens (we use polymer-starter-kit 2 here), it's first append out source custom-style and then dependency custom-style where all out styles getting overrides.

Any help would be really appreciated...

Bubbit:
I also have this issue, with vulcanize 1.14.8 it works, it seems to be broken since 1.14.10, couldn't get 1.14.9 to run due to: Error: stream.push() after EOF on same code base (with polymer-build 0.4.1 & 0.5.0)

hvdb:
The problem lays in the vulcan.js -> flatten function.

The body of a document and the head are parsed/treated differently.
While this is not a problem with generic html in general it is for css styling ofc..


We (@Bubbit and I) found a workaround.
If you put the custom style in a html file and link that inside the page where you want to override the styles, the styles will be put in the correct order.

example:

black-theme.html
```html
<link rel="import" href="../paper-styles/default-theme.html">


<style>
//some style
</style>
```

Workaround: 
Put the <style></style> part inside a new html file : b-theme.html

and import that in black-theme.html
```html
<link rel="import" href="../paper-styles/default-theme.html">

<link rel="import" href="b-thtme.html">

```


We will do some more investigation to see how we can fix it.
But there is a lot changed in this part recently so it is hard to find a good solution which doesn't break the last changes (script order)

May it is not even that bad to have the style in a different HTML file?

Bubbit:
Isn't this a duplicate of #357 ?

usergenic:
Vulcanize 1.15.1 released.  Should fix this.

PR #403, opened 2016-12-08
Style order preservation

usergenic:
vulcanize was improperly reordering styles because it moved some things out of head but not all things.  I decided the right thing to do was to only move things out of head once an html import is encountered.  I also decided that once an html import is encountered, all subsequent styles, including inlined ones, have to be moved out of head so that style ordering works right. 

 - Fixed: Preserve style ordering when moving imports and links out of head tag.
 - Don't actually move anything out of head until an html import is encountered.

### DIFF
--- a/lib/matchers.js
+++ b/lib/matchers.js
@@ -43,6 +43,15 @@ var polymerExternalStyle = p.AND(
   p.hasAttrValue('type', 'css')
 );
 
+var htmlImport = p.AND(
+  p.hasTagName('link'),
+  p.hasAttrValue('rel', 'import'),
+  p.OR(
+    p.NOT(p.hasAttr('type')),
+    p.hasAttrValue('type', 'text/html')
+  )
+);
+
 var styleMatcher = p.AND(
   p.hasTagName('style'),
   p.OR(
@@ -80,6 +89,7 @@ module.exports = {
   urlAttrs: urlAttrs,
   targetMatcher: targetMatcher,
   polymerExternalStyle: polymerExternalStyle,
+  htmlImport: htmlImport,
   JS: jsMatcher,
   CSS: styleMatcher,
   CSS_LINK: externalStyle,
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -177,9 +177,13 @@ Vulcan.prototype = {
   },
 
   moveToBodyMatcher: dom5.predicates.AND(
+    dom5.predicates.NOT(
+        dom5.predicates.parentMatches(
+            dom5.predicates.hasTagName('template'))),
     dom5.predicates.OR(
       dom5.predicates.hasTagName('script'),
-      dom5.predicates.hasTagName('link')
+      dom5.predicates.hasTagName('link'),
+      matchers.CSS
     ),
     dom5.predicates.NOT(
       dom5.predicates.OR(
@@ -215,6 +219,9 @@ Vulcan.prototype = {
     return false;
   },
 
+  isInsideTemplate: dom5.predicates.parentMatches(
+      dom5.predicates.hasTagName('template')),
+
   flatten: function flatten(tree, isMainDoc) {
     var doc = tree.html.ast;
     var imports = tree.imports;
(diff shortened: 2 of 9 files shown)


usergenic:
This should fix #357 and #375

lib/matchers.js
```diff
@@ -43,6 +43,15 @@ var polymerExternalStyle = p.AND(
   p.hasAttrValue('type', 'css')
 );
 
+var htmlImport = p.AND(
```
justinfagnani: we really need to consolidate these somewhere. I think some other projects don't have the `type` attribute check.

lib/vulcan.js
```diff
@@ -177,9 +177,13 @@ Vulcan.prototype = {
   },
 
   moveToBodyMatcher: dom5.predicates.AND(
+    dom5.predicates.NOT(
```
justinfagnani: what caused this to be needed? `<template>`s in head?

lib/vulcan.js
```diff
@@ -177,9 +177,13 @@ Vulcan.prototype = {
...
+        dom5.predicates.parentMatches(
+            dom5.predicates.hasTagName('template'))),
     dom5.predicates.OR(
       dom5.predicates.hasTagName('script'),
-      dom5.predicates.hasTagName('link')
+      dom5.predicates.hasTagName('link'),
```
justinfagnani: Is this right? I think there might be links and scripts we don't want to move: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

Maybe we only move import and stylesheet?

lib/vulcan.js
```diff
@@ -215,6 +219,9 @@ Vulcan.prototype = {
     return false;
   },
 
+  isInsideTemplate: dom5.predicates.parentMatches(
```
justinfagnani: is `parentMatches` sufficient? DOes this match ancestors?

lib/vulcan.js
usergenic: the OR includes the various attribute states for the script and link types we don't want to move.  I just added style because we do want to move those too.

lib/vulcan.js
usergenic: So when vulcanize was called on this file (test/html/inside-template.html)...
```
<!doctype html>
<!--
    @license
    Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
    This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
    The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
    The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
    Code distributed by Google as part of the polymer project is also
    subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
-->
<template>
  <link rel="import" href="external.html">
</template>
```

It bundles it up like this:
```
<!DOCTYPE html><!--
    @license
    Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
    This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
    The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
    The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
    Code distributed by Google as part of the polymer project is also
    subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
--><html><head><meta charset="UTF-8"><template>
  <link rel="import" href="external.html">
</template>
</head><body></body></html>
```

lib/matchers.js
usergenic: I had to put it there to exclude the polymer stylesheet case.

usergenic:
@justinfagnani all comments replied; PTAL

lib/vulcan.js
justinfagnani: +100

test/test.js
```diff
@@ -1004,6 +1004,28 @@ suite('Vulcan', function() {
       };
       process('test/html/inline-styles.html', callback, options);
     });
+
+    test('Inlined styles have original order preserved', function(done) {
+      var callback = function(err, doc) {
+        if (err) {
+          return done(err);
+        }
+        //assert.equal(dom5.serialize(doc), '');
```
justinfagnani: remove?

usergenic: Thank you. Removed.

PR #408, opened 2017-01-10
Fix issue with path rewriting, (especially assetpath), for certain import path combinations

usergenic:

### DIFF
--- a/lib/pathresolver.js
+++ b/lib/pathresolver.js
@@ -39,6 +39,9 @@ PathResolver.prototype = {
       for (var j = 0, attr; j < constants.URL_ATTR.length; j++) {
         attr = constants.URL_ATTR[j];
         attrValue = dom5.getAttribute(node, attr);
+        if (attrValue === '') {
+          attrValue = '.';
+        }
         if (attrValue && !this.isTemplatedUrl(attrValue)) {
           var relUrl;
           if (attr === 'style') {
--- a/lib/pathresolver.js
+++ b/lib/pathresolver.js
@@ -39,6 +39,10 @@ PathResolver.prototype = {
       for (var j = 0, attr; j < constants.URL_ATTR.length; j++) {
         attr = constants.URL_ATTR[j];
         attrValue = dom5.getAttribute(node, attr);
+        // When the path is an empty string value the intent is to represent
+        // in URL-speak "the same path that the current document is at".
+        // We'll set it to '.' which better represents this intent to the
+        // rewrite methods.
         if (attrValue === '') {
           attrValue = '.';
         }
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -222,8 +222,11 @@ Vulcan.prototype = {
   isInsideTemplate: dom5.predicates.parentMatches(
       dom5.predicates.hasTagName('template')),
 
-  flatten: function flatten(tree, rootDocPath) {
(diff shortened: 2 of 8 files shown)


usergenic:
Note: This doesn't need two reviewers, I just requested to either Fred or Peter in case one could grab earlier so we could release fix sooner.

lib/pathresolver.js
```diff
@@ -39,6 +39,9 @@ PathResolver.prototype = {
       for (var j = 0, attr; j < constants.URL_ATTR.length; j++) {
         attr = constants.URL_ATTR[j];
         attrValue = dom5.getAttribute(node, attr);
+        if (attrValue === '') {
```
FredKSchott: Can you explain what is happening here a bit more? When is `attrValue` = `''`? If that is something that bundler is causing, Isn't that the real bug that needs to be fixed?

usergenic: This is actually valid assetpath if imported doc is in same folder.  The problem is the logic that checks whether to rewrite it is using truthiness of the value of empty string both here and in the two rewrite functions.  The '.' value better represents the relative path needed in the rewrite functions, so I set it here to solve all three uses.  However, setting assetpath itself to '.' is not desirable.

usergenic: I'll add a comment :)

usergenic:
Just in case travis logs go away, the tests prove this was what the result is before the fix:

```
  1) Vulcan Regression Testing Assetpath rewriting:
      AssertionError: expected [ Array(4) ] to deeply equal [ Array(4) ]
      + expected - actual
       [
         [
           "test-c"
      -    ""
      +    "../../bower_components/test-component/"
         ]
         [
           "test-b"
           "../../bower_components/test-component/src/elements/"
      
      at Function.assert.deepEqual (node_modules/chai/lib/chai/interface/assert.js:216:32)
      at test/test.js:1137:16
      at test/test.js:308:7
      at lib/vulcan.js:510:7
```

lib/vulcan.js
```diff
@@ -222,7 +222,8 @@ Vulcan.prototype = {
   isInsideTemplate: dom5.predicates.parentMatches(
       dom5.predicates.hasTagName('template')),
 
-  flatten: function flatten(tree, isMainDoc) {
+  flatten: function flatten(tree, rootDocPath) {
+    var isMainDoc = tree.href === rootDocPath;
```
rictic: Are both `tree.href` and `rootDocPath` normalized at this point such that equality is guaranteed for equivalent paths?

usergenic: The only time this will equal is when tree.href is actually provided as the rootDocPath.  Perhaps a different way to write this is to not provide the value and test for its absence and then set it in the method like:

```
if (rootDocPath === undefined) {
  var isMainDoc = true;
  rootDocPath = tree.href;
}
```

and then call flatten with just `flatten(tree)` but I think what I have reads a little better or at least I'm ambivalent about the distinction.

usergenic: But perhaps the question is more about the behavior of the code as opposed to style.  Behavior is guaranteed because I pass in tree.href in the first instance.

usergenic:
Per rictic's query and my reflection I made `flatten` call a little more intention revealing.  Also I got rid of variable named rootDocPath in favor of the mainDocUrl convention that was already present elsewhere.  This ties it together better.

PR #413, opened 2017-01-14
Rejigger for build

usergenic:

Nothing meaningful to add to the CHANGELOG.  I just modified the way things are exposed and filter the actual npm package contents so they don't include src etc.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -79,7 +79,7 @@ export interface Options {
   stripExcludes?: string[];
 }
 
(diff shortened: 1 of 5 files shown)


src/bundler.ts
```diff
@@ -79,7 +79,7 @@ export interface Options {
   stripExcludes?: string[];
 }
 
-class Bundler {
+export class Bundler {
```
rictic: +1

Issue #412, opened 2017-01-13
Everything (even markup) ends up in hidden div


mikaelharsjo:
## Given this input html
```html
<html>
  <body>
    <h1>Test</h1>
  </body>
</html>
```

`polymer-builder index.html`

produces
```html
<html><head></head><body><div hidden="" by-polymer-bundler="">

    <h1>Test</h1>


</div></body></html>
```

while `vulcanize index.html` used to produce 

```html
<html><head><meta charset="UTF-8"></head><body>
    <h1>Test</h1>
  

</body></html>
```

The content is then ofc not visible because of style sheet.
Tested with  `v2.0.0-pre.1` and `1.14.12`

usergenic:
So this issue is fixed.  There's a related issue #415 that's addressed in #416 but the problem described was fixed in #414.

PR #414, opened 2017-01-20
Don't wrap the entrypoint doc contents in the hidden div

usergenic:
This should fix the worst of the symptoms of #412.  Bundling still synthesizes the output document, but the solution here was to modify the inlining code so that it handled the entrypoint doc differently when importing its contents.

I tried a different tack before this, which was to not synthesize a new document, but rather, append imports etc into the entrypoint document.  That turned out to be fraught with issues and I think this solution is far cleaner.

 - [ ] CHANGELOG.md has been updated

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -310,8 +310,6 @@ export class Bundler {
       reachedImports.add(resolvedUrl);
       this.rewriteImportedUrls(importDoc, resolvedUrl, docUrl);
 
-      // TODO(usergenic): remove the remove() call when PolymerLabs/dom5#35 is
-      // fixed
       let hiddenDiv = dom5.query(body, matchers.hiddenDiv);
       if (!hiddenDiv) {
         hiddenDiv = this.createHiddenContainerNode();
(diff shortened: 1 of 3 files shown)


src/bundler.ts
```diff
@@ -300,37 +310,53 @@ export class Bundler {
       reachedImports.add(resolvedUrl);
       this.rewriteImportedUrls(importDoc, resolvedUrl, docUrl);
 
-      let importParent: ASTNode;
       // TODO(usergenic): remove the remove() call when PolymerLabs/dom5#35 is
```
FredKSchott: looks like there was an update to #35, still needed? https://github.com/Polymer/dom5/issues/35

src/bundler.ts
```diff
@@ -300,37 +310,53 @@ export class Bundler {
...
-       * inline them.
-       *
-       * Finding them first saves on deduplication, moving them before inlining
-       * them
-       * allows the hidden div to always be found.
-       */
       const nestedImports = dom5.queryAll(importDoc, matchers.htmlImport);
-      astUtils.insertAllBefore(importParent, htmlImport, importDoc.childNodes!);
+
+      // If we're not hiding the import, because we're building the bundle
```
FredKSchott: nit: these two comments were a little hard to follow/understand. Would love if they could be cleaned up a bit

src/bundler.ts
```diff
@@ -548,12 +571,6 @@ export class Bundler {
         this._appendImport(hiddenDiv, newUrl);
       }
     }
-
-    /**
-     * If there's only one entrypoint, get all of its DOM into the synthetic
-     * import.
-     */
```
usergenic: This comment was meaningless and didn't even relate to the code below it.

src/bundler.ts
```diff
@@ -284,6 +291,9 @@ export class Bundler {
       return;
     }
 
+    const doc = dom5.nodeWalkAncestors(htmlImport, (node) => !node.parentNode)!;
+    const body = dom5.query(doc, dom5.predicates.hasTagName('body'))!;
```
usergenic: There was some weird and fiddly code in the beforeHiddenDiv/afterHiddenDiv/inHiddenDiv logic that also found the hiddenDiv in different ways.  It is simpler to just query the doc body for it, which is why we get body here.

src/bundler.ts
usergenic: Good point.  Removed comment and removed 3 unneeded remove calls.

usergenic:
Discovered as a result of this work is the issue #415 which I'll solve in a subsequent PR.

Issue #402, opened 2016-12-08
addedImports/--add-import is broken


paulie4:
The [documentation ](https://github.com/Polymer/vulcanize#using-vulcanize-programmatically) says this:
> addedImports: Additional HTML imports to inline, added to the end of the target file

but it is doing the complete opposite. It is not inlining the import and instead is adding a `<link rel=...>` for the `addedImports` to the start of `<head>`.

usergenic:
This is a good catch, @paulie4 - Even the test proves vulcanize is just adding a `<link href="..."` to the doc.

usergenic:
Not sure if documentation wrong or code misimplemented.   The `--add-import` flag for example describes it as "Add this import to the target HTML before vulcanizing" which implies it would be inlined, but doesn't expressly state that.

I'll mark this as a bug.

PR #416, opened 2017-01-26
Mass refactor

### LINKED ISSUE #412: Everything (even markup) ends up in hidden div

### LINKED ISSUE #415: Tags that belong in `<head>` of the bundle target are erroneously moved into `<body>`.
A by-product of the choice to synthesize bundle files instead of injecting content into entrypoint files was an oversimplification of the inlining code, such that everything *was* bundled into a hidden div.  That's fixed in #414 but there are tags that only make sense in `<head>` that were moved into `<body>` but we don't want to do that.  Need to add special handling code to ensure as many tags from `<head>` of bundle doc stay in (are inlined back into) `<head>`

usergenic:
Okay.

So this all started when I wanted to fix #415.  The problem that surfaced investigating solutions to that bug was actually based in the design choice to synthesize bundle files from scratch and inject the various bits into them, which meant that bundler didn't produce a `vulcanize` equivalent and broke assumptions about the generated files.

What was done was do say, lets get rid of `_synthesizeBundleDocument` and instead `_prepareBundleDocument` from an available source doc (if bundle url resolved to one) OR an empty doc (this would be for synthesized common libraries) and afterwards just run an `_appendHtmlImports` loop to inject the import directives into whatever was prepared.

In doing this, I touched a lot of code and it became clear the more I touched, the more could be completely pulled out or refactored, so rather than doing this piecemeal, I took a bit of a sledgehammer to it and made a mongo PR.

Given the sweeping nature of the PR, I did the following to make review a little smoother:

 - Added comments for every function that didn't have them
 - Opportunistically moved the more static behaving functions out of bundler.ts into import-utils.ts, though more could probably be moved.
 - Removed unused code mercilessly, which actually allowed me to identify a few latent untested bugs and also fix them opportunistically.
 - Made everything that wasn't intended to be a public API on the Bundler class private for now.  This highlights the interface better.
 - Sorted methods alphabetically on Bundler, because they were just all over the place for no good reason.

I think this code can still get a lot simpler than it is, but this at least brings it into a condition where it could be reworked more iteratively.

Note this solves for #415 and while it removes code that fixed #412, it obviates the need for that fix with the new bundle document strategy.


src/ast-utils.ts
```diff
@@ -94,3 +92,62 @@ export function querySelectorAllTemplates(
   }
   return results;
 }
+
+/**
+ * The results of `queryAll` combined with `querySelectorAllTemplates`.
+ */
+export function querySelectorAllWithTemplates(
```
justinfagnani: dom5 now supports optionally walking into templates, use that?

src/bundler.ts
```diff
@@ -46,6 +47,7 @@ export interface Options {
   // or other hint to fix the top-level folder.
   basePath?: string;
 
+  // TODO(usergenic): Added Imports is not yet supported.
```
justinfagnani: can this be removed for now then?

src/bundler.ts
```diff
@@ -112,149 +114,185 @@ export class Bundler {
...
+   * documents with HTML imports, external stylesheets and external scripts
+   * inlined according to the options for this Bundler.
    *
-   * If the URL is part of another bundle, this method returns the url of that
-   * bundle.
-   *
-   * If the URL isn't part of a bundle, this method returns `false`
+   * @param {Array<string>} entrypoints The list of entrypoints that will be
+   *     analyzed for dependencies. The results of the analysis will be passed
+   *     to the `strategy`. An array of length 1 will bypass the strategy and
```
justinfagnani: Interesting... I'm not sure we want to bypass the strategy ultimately, because with lazy imports if something is only ever imported lazily, it'll act like an entrypoint.

src/bundler.ts
```diff
@@ -112,149 +114,185 @@ export class Bundler {
...
+   *     directly bundle the document.
+   * @param {BundleStrategy} strategy The strategy used to construct the
+   *     output bundles. See 'polymer-analyzer/lib/bundle-manifest'.
```
justinfagnani: is this path correct? not `src/bundle-manifest.ts`?

src/ast-utils.ts
```diff
@@ -17,66 +17,64 @@ import {ASTNode, treeAdapters} from 'parse5';
...
+ * Inserts the node as the first child of the parent.
  */
-export function querySelectorAllWithTemplates(
-    node: ASTNode, predicate: dom5.Predicate, noRecursion?: boolean):
-    ASTNode[] {
-  let results = dom5.queryAll(node, predicate);
-  results =
-      results.concat(querySelectorAllTemplates(node, predicate, noRecursion));
-  return results;
+export function prepend(parent: ASTNode, node: ASTNode) {
```
usergenic: TODO added.

src/ast-utils.ts
```diff
@@ -17,66 +17,64 @@ import {ASTNode, treeAdapters} from 'parse5';
...
-  const index = before.parentNode!.childNodes!.indexOf(before);
-  before.parentNode!.childNodes!.splice(index + 1, 0, node);
-  node.parentNode = before.parentNode!;
+/**
+ * Return true if node is a text node that is empty or consists only of white
+ * space.
+ */
+export function isBlankTextNode(node: ASTNode): boolean {
+  return node && dom5.isTextNode(node) &&
+      /^\s*$/.test(dom5.getTextContent(node));
```
usergenic: Done.  That reads simpler.

src/ast-utils.ts
```diff
@@ -17,66 +17,64 @@ import {ASTNode, treeAdapters} from 'parse5';
...
-  while (moveIndex >= 0) {
-    const nodeToMove = nodes[moveIndex];
-    dom5.remove(nodeToMove);
-    prepend(parent, nodeToMove);
-    moveIndex--;
-  }
+/**
+ * Move the `node` to be the immediate sibling after the `before` node.
+ */
+export function insertAfter(before: ASTNode, node: ASTNode) {
```
usergenic: Added TODO.

usergenic: Renamed before to target.

src/bundler.ts
usergenic: Great catch.  The comment is wrong actually.  I culled the code that branched based on entrypoint length but didn't see this point in the comment.  Comment fixed.

src/bundler.ts
usergenic: That's a good point.  If someone wants to *see* they should visit src instead of lib.  Fixed comments.

src/ast-utils.ts
usergenic: Oh look at that!  Didn't know that was a thing, `dom5.childNodesIncludeTemplate`.  Done!

src/bundler.ts
usergenic: I only want to remove this if it's off the feature list for bundler, since it's planned to bring this forward to maintain parity with vulcanize.

usergenic: Hmm.  I say that, but now looking at what vulcanize is actually doing with it per this issue #402 and checking out the test-- The behavior of addedImports seems counter to what one would expect.  The feature was added to vulcanize in #248 but no discussion as to reason why feature was added is given or why it was implemented without inlining.  We should discuss.

src/import-utils.ts
```diff
@@ -0,0 +1,300 @@
...
+import * as parse5 from 'parse5';
+import {ASTNode} from 'parse5';
+import {Analyzer, Options as AnalyzerOptions} from 'polymer-analyzer';
+import {Document, ScannedDocument} from 'polymer-analyzer/lib/model/document';
+import {Import} from 'polymer-analyzer/lib/model/import';
+import {ParsedHtmlDocument} from 'polymer-analyzer/lib/html/html-document';
+import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
+import constants from './constants';
+import * as astUtils from './ast-utils';
+import * as importUtils from './import-utils';
```
FredKSchott: womp womp

usergenic:
FYI, @FredKSchott and I just sat down together to do a big in-person review of the code.  I'll be pushing commits to address each item as I fix them and follow those up with a final comment.

src/bundler.ts
usergenic: And upon further further review it is clear that as defined, the `--add-import` option has no means to correctly relate to which bundle the added import would even be attached.  I've removed the option for now because:

 1. There's no clear way to actually implement it as expressed with current option structure.
 2. The result of adding import is surprising and contrasts with intended behavior of bundling.

justinfagnani:
I'm trying to comment about removing `--add-import` but can't find it. It's the right decision though. +1

usergenic:
Okay.  Have addressed all of the in-person feedback from @FredKSchott.  PTAL @justinfagnani

src/bundler.ts
```diff
@@ -102,346 +97,268 @@ export class Bundler {
...
     dom5.setAttribute(hidden, 'hidden', '');
     dom5.setAttribute(hidden, 'by-polymer-bundler', '');
     return hidden;
   }
 
   /**
-   * Inline external scripts <script src="*">
+   * Given an array of Bundles, remove all files from bundles which are in the
+   * "excludes" set.  Remove any bundles which are left empty after excluded
+   * files are removed.
```
justinfagnani: Maybe note that while you mutate the Bundles, you don't mutate the bundles array. Would eventually be nice if this were a pure function, maybe if Bundle had a copy constructor.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+  dom5.removeAttribute(externalScript, 'src');
+  dom5.setTextContent(externalScript, scriptContent);
+  return scriptContent;
+}
+
+/**
+ * Inlines the contents of the stylesheet returned by the link tag's href url
+ * into a style tag and removes the link tag.
+ */
+export async function inlineStylesheet(
```
justinfagnani: Huh, now I see that you moved much of this out of the Bundler. That might be making some things more cumbersome just because you don't have all the state on an instance anymore, it's all passing as arguments.

src/ast-utils.ts
```diff
@@ -17,80 +17,106 @@ import {ASTNode, treeAdapters} from 'parse5';
...
-    moveIndex--;
-  }
+/**
+ * Move the `node` to be the immediate sibling after the `target` node.
+ * TODO(usergenic): Migrate this code to polymer/dom5
+ */
+export function insertAfter(target: ASTNode, node: ASTNode) {
+  dom5.remove(node);
+  const index = target.parentNode!.childNodes!.indexOf(target);
+  target.parentNode!.childNodes!.splice(index + 1, 0, node);
```
usergenic: Meh dom5 doesn't export insertNode today.  I updated the TODO comment on this function to make that change when porting it to dom5.

src/bin/polymer-bundler.ts
```diff
@@ -194,7 +186,7 @@ function stringToRegExp(str: string) {
   return new RegExp(str.replace(/[-\/\\*+?.()|[\]{}]/g, '\\$&'));
 }
 
-options.addedImports = options['add-import'] || [];
+options.basePath = options['abspath'];
```
usergenic: Yeah I think the intent was backwards compatibility.  I've renamed it `--base-path` to correspond to the Bundler constructor option of `basePath`.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+ * Inline the contents of the html document returned by the link tag's href
+ * at the location of the link tag and then remove the link tag.
+ */
+export async function inlineHtmlImport(
+    basePath: string|undefined,
+    docUrl: string,
+    htmlImport: ASTNode,
+    reachedImports: Set<string>,
```
usergenic: Renamed to visitedUrls

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+    docBundle: AssignedBundle,
+    manifest: BundleManifest,
+    loader: (url: UrlString) => Promise<string>) {
+  const rawImportUrl: string = dom5.getAttribute(htmlImport, 'href')!;
```
usergenic: Removed type annotations for assignments.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+  const resolvedImportUrl: string = urlLib.resolve(docUrl, rawImportUrl);
+  const importBundleUrl = manifest.bundleUrlForFile.get(resolvedImportUrl);
+
+  // Don't reprocess the same file again.
+  if (reachedImports.has(resolvedImportUrl)) {
+    astUtils.removeElementAndNewline(htmlImport);
+    return;
+  }
```
usergenic: Fair enough.  Done'd.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+
+  // Don't inline an import into itself.
+  if (docUrl === resolvedImportUrl) {
+    reachedImports.add(resolvedImportUrl);
+    astUtils.removeElementAndNewline(htmlImport);
+    return;
+  }
+
+  // Guard against inlining a import we've already processed.
+  if (reachedImports.has(importBundleUrl)) {
```
usergenic: Updated teh comment:

```

  // If we've previously visited a url that is part of a bundle, it means we've
  // handled that entire bundle, so we guard against inlining any other file
  // from that bundle by checking the visited urls for the bundle url itself.
  if (visitedUrls.has(importBundleUrl)) {
    astUtils.removeElementAndNewline(htmlImport);
    return;
  }

  // If the html import refers to a file which is bundled and has a different
  // url, then lets just rewrite the href to point to the bundle url.
  if (importBundleUrl !== docBundle.url) {
    const relative =
        urlUtils.relativeUrl(docUrl, importBundleUrl) || importBundleUrl;
```

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+  // url, then lets just rewrite the href to point to the bundle url.
+  if (importBundleUrl !== docBundle.url) {
+    const relative =
+        urlUtils.relativeUrl(docUrl, importBundleUrl) || importBundleUrl;
+    dom5.setAttribute(htmlImport, 'href', relative);
+    reachedImports.add(importBundleUrl);
+    return;
+  }
+
+  const document =
```
usergenic: It actually turns out I don't use `document` or `body` here anymore anyways. I am going to update bundler's tslint configuration to catch unused variables.

usergenic: Added `no-unused-variable: true` in 01403af55fc28482904e91fb5180bbdaa4c33767 and it caught a lot of things.  Not as many as I expected, but a few unused functions and such were pruned.  Lots of unused imports too of course.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+      dom5.nodeWalkAncestors(htmlImport, (node) => !node.parentNode)!;
+  const body = dom5.query(document, matchers.body)!;
+  const importSource = await loader(resolvedImportUrl).catch(err => {
+    throw new Error(`Unable to load ${resolvedImportUrl}: ${err.message}`);
+  });
+
+  // Is there a better way to get what we want other than using
+  // parseFragment?
```
usergenic: That was an old comment from when AJ and I were pairing.  I got rid of it.  You answered the question anyways :)  Thanks!

src/import-utils.ts
usergenic: It might look like I'm passing around state, but I'm not really, or at least, I'm not passing around state of an object that's been yet been defined.  The passed arguments to these functions are specific to preparing a single bundled document and include the details necessary to add or do one thing to it as part of that process.  I could could name that state and build a class, but I don't have a great name for it yet.  I'll could add a TODO to the top of import-utils.ts to revisit and possibly refactor into a class called DocumentBuilder or something to help simplify the function calls, but ultimately these are functions that manipulate the dom by-way-of the elements they're handed and they don't change state of anything else, so I'm a little ambivalent.  I do agree that inlineHtmlImport has a truckload of arguments and I'd like to reduce that because eyesore.

src/bundler.ts
```diff
@@ -26,28 +26,28 @@ import {ParsedHtmlDocument} from 'polymer-analyzer/lib/html/html-document';
...
-import {Bundle, BundleStrategy, AssignedBundle, generateBundles, BundleUrlMapper, BundleManifest, sharedBundleUrlMapper, generateSharedDepsMergeStrategy} from './bundle-manifest';
+import * as bundleManifestLib from './bundle-manifest';
+import {Bundle, BundleStrategy, AssignedBundle, BundleUrlMapper, BundleManifest} from './bundle-manifest';
 import {BundledDocument, DocumentCollection} from './document-collection';
-import {buildDepsIndex} from './deps-index';
+import * as depsIndexLib from './deps-index';
 import {UrlString} from './url-utils';
 
-// TODO(usergenic): Document every one of these options.
+// TODO(garlicnation): resolve <base> tags.
```
usergenic: K

src/bundler.ts
```diff
@@ -26,28 +26,28 @@ import {ParsedHtmlDocument} from 'polymer-analyzer/lib/html/html-document';
...
   // When provided, relative paths will be converted to absolute paths where
   // `basePath` is the root url.  This path is equal to the folder of the
   // bundled url document of the analyzer.
-  //
-  // TODO(usergenic): If multiple `bundle()` calls are made `basePath` can
-  // produce incompatile absolute paths if the same `basePath` is used for
-  // `bundle()` targets in different folders.  Possible solutions include
-  // removing basePath behavior altogether or supplementing it with a `rootPath`
-  // or other hint to fix the top-level folder.
   basePath?: string;
```
usergenic: @justinfagnani Are you saying that there's another something down the toolchain which can prefix all the paths in all the the things and we can pull this out of bundler altogether because I WOULD LOVE THAT.

justinfagnani: Exactly. That thing could come before or after the bundler.

usergenic: basePath is dead. f0455c72b6c07958b731be0294f39e0f201def74 !!!

src/bundler.ts
```diff
@@ -102,346 +97,268 @@ export class Bundler {
...
-    return true;
+
+    return bundledDocuments;
   }
 
-  isStripExcludedHref(url: string): boolean {
-    if (!this.stripExcludes) {
-      return false;
+  /**
+   * Add HTML Import elements for each file in the bundle.  We append all the
```
usergenic: So, we visit all of the link tags we encounter while walking the imports recursively etc.  This is true.  And most of the time that will include all the imports we'd care about.  However, the inlining logic is still driven by the bundle manifest and decision whether to inline content or link to a bundle is decided by the manifest.

Additionally, during the definition of bundles, a strategy might decide to move files somewhere else altogether, at which time, we can't rely on the link tags actually existing in a source to signal they should be inlined into that bundle, so we brute force the problem here by just appending links to everything we know a bundle should have into the div and relying on the visitedUrls logic to effectively deduplicate the lot, while allowing the expected imports we would not have otherwise visited to be found and inlined.

usergenic: This will most commonly be because we're creating a shared bundle for common libs, which has no single source document containing all those links.  But the rule works even if we have an entrypoint document that a custom bundle strategy decided to add an import to.

src/bundler.ts
```diff
@@ -102,346 +97,268 @@ export class Bundler {
...
+      bundle: AssignedBundle) {
+    for (const importUrl of bundle.bundle.files) {
+      const newUrl = urlUtils.relativeUrl(bundle.url, importUrl);
+      if (!newUrl) {
+        continue;
+      }
+      this._appendHtmlImport(this._findOrCreateHiddenDiv(document), newUrl);
     }
-    return this.stripExcludes.some(r => url.search(r) >= 0);
+    return document;
```
usergenic: 1. It wasn't used.
2. It doesn't even match the type signature for the function.
3. I've removed it.
4. Thanks!

src/bundler.ts
```diff
@@ -102,346 +97,268 @@ export class Bundler {
...
+  /**
+   * Append a <link rel="import" node to `node` with a value of `url` for
+   * the "href" attribute.
+   */
+  private _appendHtmlImport(node: ASTNode, url: UrlString): ASTNode {
+    const newNode = dom5.constructors.element('link');
+    dom5.setAttribute(newNode, 'rel', 'import');
+    dom5.setAttribute(newNode, 'href', url);
+    dom5.append(node, newNode);
+    return newNode;
```
usergenic: Nah.  It *could* have been used, but since we don't use it, I'll get rid of it.

src/bundler.ts
usergenic: This bothered you too, huh?

I've changed the function to not return and just mutate array in-place.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+import constants from './constants';
+import * as astUtils from './ast-utils';
+import * as matchers from './matchers';
+import * as urlUtils from './url-utils';
+import {Bundle, BundleStrategy, AssignedBundle, generateBundles, BundleUrlMapper, BundleManifest, sharedBundleUrlMapper, generateSharedDepsMergeStrategy} from './bundle-manifest';
+import {BundledDocument, DocumentCollection} from './document-collection';
+import {buildDepsIndex} from './deps-index';
+import {UrlString} from './url-utils';
+
+// TODO(usergenic): Want to figure out a way to get rid of the basePath param
```
usergenic: Getting rid of basePath helped.  Anyways this comment is changed now to speculate about a class for the import utils to deal with state/shared params etc.

src/bin/polymer-bundler.ts
usergenic: and... it's gone.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+        urlUtils.relativeUrl(docUrl, importBundleUrl) || importBundleUrl;
+    dom5.setAttribute(htmlImport, 'href', relative);
+    reachedImports.add(importBundleUrl);
+    return;
+  }
+
+  const document =
+      dom5.nodeWalkAncestors(htmlImport, (node) => !node.parentNode)!;
+  const body = dom5.query(document, matchers.body)!;
+  const importSource = await loader(resolvedImportUrl).catch(err => {
```
usergenic: good *catch*.
fixed

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+ * Inlines the contents of the document returned by the script tag's src url
+ * into the script tag content and removes the src attribute.
+ */
+export async function inlineScript(
+    docUrl: UrlString,
+    externalScript: ASTNode,
+    loader: (url: UrlString) => Promise<string>) {
+  const rawUrl: string = dom5.getAttribute(externalScript, 'src')!;
```
usergenic: Undid it.

src/import-utils.ts
```diff
@@ -0,0 +1,301 @@
...
+  const resolvedUrl = urlLib.resolve(docUrl, rawUrl);
+  let script: string|undefined = undefined;
+  try {
+    script = await loader(resolvedUrl);
+  } catch (err) {
+    // If a script doesn't load, skip inlining.
+    // TODO(usergenic): Add plylog logging for load error.
```
usergenic: I added the console.warn but now my test output has warnings.

😿

usergenic:
@justinfagnani addressed all comments.  PTAL!

src/bundler.ts
justinfagnani: +!

src/bundler.ts
justinfagnani: Seems like we need to be careful to define what happens when we have a graph of imports in documents from an existing root of a bundle, and a new authoritative set of document that the bundle contains from the manifest. Appending imports would be one way of implementing   that. Another bundler would want to match the behavior even if they didn't do it by appending imports.

The important aspect of that is ordering. It seems like this approach preserves an important property - that the order of imports in actual files is respected first, and only if an import isn't already transitively reachable from an existing bundle root is it's ordering determined by the manifest.

I would slightly and eventually like to see this done without injecting imports, since imports that aren't part of the bundle are moved on the fly - seems like additional imports should be able to be added after inlining everything and seeing what's left. But that's a very low priority wish.

src/import-utils.ts
justinfagnani: ack

src/ast-utils.ts
justinfagnani: dom5 needs an overhaul based on actual usage.

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
...
+- Changed the way bundle documents are generated, so that bundles based on a
+  source file retain their HTML structure instead of always being synthesized
+  from a blank document.
+- bundle() is the only public method on Bundler class now.
+- Bundler does not explicitly throw an error if it spots a `<polymer-element>`
+  somewhere in a document.
+- `--add-import` and `--importUrl` options have been removed.  These features,
+  as expressed, need to be conceptually reworked to work in bundler's new
+  multi-bundle paradigm.
+- `--abspath`/`basePath` have been removed, anticipating their reimplementation
```
justinfagnani: did you file an issue on this? If so, this is a good place to link it

usergenic: Yeah filed-- Added link in changelog.  Good idea!

src/bundler.ts
usergenic: Yeah we could easily see using the visitedUrls set to create an exclusion set for the remaining files to inline but also meh.  We might do this if we add a feature that can return a manifest of all inlined things, because then we'd have code to track the importing that would need to preserve state through these various inlining calls etc.

PR #421, opened 2017-02-01
Fix shell bundle strategy

usergenic:
Fixed a small bug where the generateShellMergeStrategy was not honoring minEntrypoints and if it wasn't passed in as a 2 it produced incorrect bundles and sometimes it also generated incorrect bundles anyways if the shared bundle happened to come after the shell bundle when the .find method was invoked on bundles set.

### DIFF
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -213,7 +213,7 @@ export function generateShellMergeStrategy(
     }
     const sharedBundles = newBundles.filter(
         (bundle) => bundle.entrypoints.size >= minEntrypoints &&
-            bundle !== shellBundle);
+            bundle !== shellBundle && !getBundleEntrypoint(bundle));
     for (const sharedBundle of sharedBundles) {
       newBundles.splice(newBundles.indexOf(shellBundle), 1);
       newBundles.splice(newBundles.indexOf(sharedBundle), 1);
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -201,16 +201,21 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
  * bundle for an application shell.
  */
 export function generateShellMergeStrategy(
-    shell: UrlString, minEntrypoints: number): BundleStrategy {
+    shell: UrlString, minEntrypoints?: number): BundleStrategy {
   return (bundles: Bundle[]): Bundle[] => {
+    if (minEntrypoints === undefined) {
+      minEntrypoints = 2;
+    }
     const newBundles = generateSharedDepsMergeStrategy(minEntrypoints)(bundles);
-    const shellBundle = newBundles.find((bundle) => bundle.files.has(shell));
-    const sharedBundle =
-        newBundles.find((bundle) => bundle.entrypoints.size > 1);
-    if (shellBundle && sharedBundle && shellBundle !== sharedBundle) {
-      newBundles.splice(newBundles.indexOf(shellBundle), 1);
-      newBundles.splice(newBundles.indexOf(sharedBundle), 1);
-      newBundles.push(mergeBundles([shellBundle, sharedBundle]));
+    const sharedBundles = newBundles.filter(
+        (bundle) => bundle.entrypoints.size >= minEntrypoints);
+    for (const sharedBundle of sharedBundles) {
+      const shellBundle = newBundles.find((bundle) => bundle.files.has(shell));
+      if (shellBundle && sharedBundle && shellBundle !== sharedBundle) {
+        newBundles.splice(newBundles.indexOf(shellBundle), 1);
+        newBundles.splice(newBundles.indexOf(sharedBundle), 1);
+        newBundles.push(mergeBundles([shellBundle, sharedBundle]));
+      }
     }
     return newBundles;
   };
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -207,15 +207,18 @@ export function generateShellMergeStrategy(
       minEntrypoints = 2;
     }
     const newBundles = generateSharedDepsMergeStrategy(minEntrypoints)(bundles);
+    let shellBundle = newBundles.find(bundle => bundle.files.has(shell));
+    if (!shellBundle) {
+      throw new Error(`No bundle found containing specified shell ${shell}`);
+    }
     const sharedBundles = newBundles.filter(
-        (bundle) => bundle.entrypoints.size >= minEntrypoints);
+        (bundle) => bundle.entrypoints.size >= minEntrypoints &&
+            bundle !== shellBundle);
     for (const sharedBundle of sharedBundles) {
-      const shellBundle = newBundles.find((bundle) => bundle.files.has(shell));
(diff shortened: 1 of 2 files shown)


src/bundle-manifest.ts
```diff
@@ -201,16 +201,21 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
...
-        newBundles.find((bundle) => bundle.entrypoints.size > 1);
-    if (shellBundle && sharedBundle && shellBundle !== sharedBundle) {
-      newBundles.splice(newBundles.indexOf(shellBundle), 1);
-      newBundles.splice(newBundles.indexOf(sharedBundle), 1);
-      newBundles.push(mergeBundles([shellBundle, sharedBundle]));
+    const sharedBundles = newBundles.filter(
+        (bundle) => bundle.entrypoints.size >= minEntrypoints);
+    for (const sharedBundle of sharedBundles) {
+      const shellBundle = newBundles.find((bundle) => bundle.files.has(shell));
+      if (shellBundle && sharedBundle && shellBundle !== sharedBundle) {
```
garlicnation: This line could use a comment explaining what the if clause is doing.

src/bundle-manifest.ts
```diff
@@ -201,16 +201,21 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
  * bundle for an application shell.
  */
 export function generateShellMergeStrategy(
-    shell: UrlString, minEntrypoints: number): BundleStrategy {
+    shell: UrlString, minEntrypoints?: number): BundleStrategy {
   return (bundles: Bundle[]): Bundle[] => {
+    if (minEntrypoints === undefined) {
```
garlicnation: This needs a test.

src/bundle-manifest.ts
```diff
@@ -201,16 +201,21 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
...
-    const sharedBundle =
-        newBundles.find((bundle) => bundle.entrypoints.size > 1);
-    if (shellBundle && sharedBundle && shellBundle !== sharedBundle) {
-      newBundles.splice(newBundles.indexOf(shellBundle), 1);
-      newBundles.splice(newBundles.indexOf(sharedBundle), 1);
-      newBundles.push(mergeBundles([shellBundle, sharedBundle]));
+    const sharedBundles = newBundles.filter(
+        (bundle) => bundle.entrypoints.size >= minEntrypoints);
+    for (const sharedBundle of sharedBundles) {
+      const shellBundle = newBundles.find((bundle) => bundle.files.has(shell));
```
garlicnation: Move out of the for loop.

usergenic: Well, can't because shellBundle changes each time through the loop.

src/bundle-manifest.ts
usergenic: Ok 😠

src/bundle-manifest.ts
usergenic: I'll restructure this using a let.  😠

src/bundle-manifest.ts
FredKSchott: +1 one on this change, but agree a test would be good. Feel free to revert this for now as unrelated and address in separate PR.

usergenic:
My 😠 faces are directed at myself fwiw. 👍

usergenic:
@garlicnation PTAL

src/bundle-manifest.ts
```diff
@@ -201,16 +201,24 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
...
+      throw new Error(`No bundle found containing specified shell ${shell}`);
+    }
+    const sharedBundles = newBundles.filter(
+        (bundle) => bundle.entrypoints.size >= minEntrypoints &&
+            bundle !== shellBundle);
+    for (const sharedBundle of sharedBundles) {
       newBundles.splice(newBundles.indexOf(shellBundle), 1);
       newBundles.splice(newBundles.indexOf(sharedBundle), 1);
-      newBundles.push(mergeBundles([shellBundle, sharedBundle]));
+      shellBundle = mergeBundles([shellBundle, sharedBundle]);
```
justinfagnani: Doesn't this imply that the value of `minEntrypoints` doesn't matter? They all get merged anyway. This will also merge all lazy bundles into the shell, yes?

justinfagnani: This is a comment to satisfy Github because my other comment wasn't part of a review.

usergenic: If you pass 3 minEntrypoints to generateShellMergeStrategy, you'll get a shell that only contains files which are depended on by at least 3 entrypoints:

```
const sharedBundles = newBundles.filter(
        (bundle) => bundle.entrypoints.size >= minEntrypoints &&
            bundle !== shellBundle);
```

usergenic: Nothing explicitly honors or disregards lazy/non-lazy bundles in any strategy code AFAIK.  Lazy bundles + shell and shared merge strategy generators are worth looking into but way beyond the scope of this.

The origin of this PR is actually just the result of noticing that finding the sharedBundle with .find() had the potential of fetching the wrong bundle if the shell was explicitly linked to by any entrypoints and based on arbitrary order of the bundles searched over.  I fixed the expectation of the behavior of sending in higher minEntrypoints values and lastly made the argument optional to make it compatible to the optional argument for generateSharedDepsMergeStrategy.  But fixing the find was the impetus.

usergenic: Github God

The Github God is satisfied.

justinfagnani: Ok, may be a false alarm, but the loop implies there may be more than one of these... and merging all the shared bundles into the shell isn't what I'd expect. But remembering the shared deps strategy, there should only be one bundle that satisfies `bundle.entrypoints.size >= minEntrypoints` I thought.

Looking at the code for `generateSharedDepsMergeStrategy` though, I'm finding it hard to reason about the `!getBundleEntrypoint(bundle)` exception for putting things in the shared bundle - that could use a comment. I think the only way that an entrypoint bundle could run have more than one entrypoint is if there were a dependency from one entrypoint to another. Either way, if there are more than 1 bundles with `bundle.entrypoints.size >= minEntrypoints`, it seems only one would also satisfy `!getBundleEntrypoint(bundle)` - should the ones that have an entrypoint be excluded from the shell bundle?

usergenic: There are definitely test cases where one entrypoint includes an html import to another entrypoint.  I can't recall if those tests were in bundler or build off-hand, but that scenario is not too far-fetched given how creative developers can be.

Hmm, but your point about the loop being misleading... I actually wrote the loop in the shell strategy code with the expectation that there might be scenarios where `generateSharedDepsMergeStrategy` should generate more than one shared bundle.  After testing it though, there don't appear to be situations where it does generate more than one.  it would I think there might be cases for it doing so in situations where there's a clear split in the dependency graph, but now I'm speculating...

usergenic:
@garlicnation and @justinfagnani PTAL WYGAC - There's nothing else to address on the PR that I'm aware of.

PR #422, opened 2017-02-01
Fix issue importing bundle

usergenic:
An issue surfaced when trying to execute the shell merge strategy in polymer-build where some imports were not making it into the shell.  I scratched my head a lot and eventually followed the problem to a logic error that came in with mass refactor that had no scenario to cover it.

Anyways, I've added a regression test for the problem scenario (a file linking to the url of its own bundle stopping import of subsequent files) and a fix.

### DIFF
--- a/src/import-utils.ts
+++ b/src/import-utils.ts
@@ -67,17 +67,18 @@ export async function inlineHtmlImport(
     return;
   }
 
-  // If we've previously visited a url that is part of a bundle, it means we've
-  // handled that entire bundle, so we guard against inlining any other file
-  // from that bundle by checking the visited urls for the bundle url itself.
-  if (visitedUrls.has(importBundleUrl)) {
-    astUtils.removeElementAndNewline(htmlImport);
-    return;
-  }
-
   // If the html import refers to a file which is bundled and has a different
   // url, then lets just rewrite the href to point to the bundle url.
   if (importBundleUrl !== docBundle.url) {
+    // If we've previously visited a url that is part of another bundle, it
+    // means we've handled that entire bundle, so we guard against inlining any
+    // other file from that bundle by checking the visited urls for the bundle
+    // url itself.
+    if (visitedUrls.has(importBundleUrl)) {
+      astUtils.removeElementAndNewline(htmlImport);
+      return;
+    }
+
     const relative =
         urlUtils.relativeUrl(docUrl, importBundleUrl) || importBundleUrl;
     dom5.setAttribute(htmlImport, 'href', relative);
--- a/src/test/shards_test.ts
+++ b/src/test/shards_test.ts
@@ -20,7 +20,7 @@ import * as parse5 from 'parse5';
 import {Analyzer} from 'polymer-analyzer';
 import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
 
-import {BundleStrategy, generateSharedDepsMergeStrategy} from '../bundle-manifest';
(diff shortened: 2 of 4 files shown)


src/import-utils.ts
```diff
@@ -67,17 +67,18 @@ export async function inlineHtmlImport(
...
   }
 
-  // If we've previously visited a url that is part of a bundle, it means we've
-  // handled that entire bundle, so we guard against inlining any other file
-  // from that bundle by checking the visited urls for the bundle url itself.
-  if (visitedUrls.has(importBundleUrl)) {
-    astUtils.removeElementAndNewline(htmlImport);
-    return;
-  }
-
```
usergenic: So I basically MOVED this block into the if below it, because it is only relevant when the importBundleUrl is not the CURRENT bundle url.

test/html/multiple-imports.html
```diff
@@ -7,5 +7,6 @@
     Code distributed by Google as part of the polymer project is also
     subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 -->
+<link rel="import" href="multiple-imports.html">
```
FredKSchott: I'm having trouble understanding the issue / when this bug appears. Does a file have to import itself? Isn't that already an error?

usergenic: Fair point.  The test needs to be more intention revealing.  Essentially, bundles with files that are not a part of the original transitive dependency chain will suffer from this problem, since imports are appended in arbitrary order to the hidden div.

In the case of running in polymer-build, what I saw happen was that the common-dependency.html was injected into the shell's bundle by the bundle strategy, because that's what it should do, and shell.html's hidden div wound up with a `<link rel=import href=shell.html><link rel=import href=common-dependency.html>` before inlining, because appendHtmlImports adds references to all files, and then the common-dependency.html wasn't never inlined because the condition that we've "already visited the importBundleUrl of shell.html" short-circuited subsequent imports.

usergenic: This is tricky to represent in a test, but the symptom of the arbitrary injection order was easy enough to replicate by adding a self-reference in the file here.  I'll see if I can clear-up the issue in tests now.

FredKSchott: Ah I think I understand now. Yea, I'd definitely like to see a clean test fixture for this case. New test fixtures are cheap, and I worry about polluting an already-in-use test with something as strange as this.

usergenic: Okay I force pushed a new version of the branch with a test that specifically addresses sharded build strategy (which fails) and subsequent commit contains the fix code (same fix code).

usergenic:
@FredKSchott PTAL thx!

src/test/shards_test.ts
```diff
@@ -97,5 +98,39 @@ suite('Bundler', () => {
                 [commonModule, elOne, depOne]);
           });
     });
+
+    test('with 2 entrypoints and a shell, all deps are in their places', () => {
+      const strategy = generateShellMergeStrategy(shell, 2);
+      return bundleMultiple([shell, entrypoint1, entrypoint2], strategy)
+          .then((docs) => {
+            //      assert.equal(docs.size, 3);
```
FredKSchott: ?

usergenic: Good catch- Thank you!  Uncommented.

usergenic:
Added a CHANGELOG note too:

```
- Fixed a bug where bundling using generateShellMergeStrategy would result in
  builds missing imports which should have been appended to shell.
```

PR #423, opened 2017-02-03
Even better strategy generators

usergenic:
@justinfagnani @garlicnation PTAL.  Here is a cleaned up set of bundle strategy generators which are composed in a (hopefully) clearer and more intention-revealing way.  I'm making this a new pull request in case for some reason we want to go with the other one instead after you look at this.

In particular, Justin and I talked this morning about how a shell strategy might just care about its own dependencies-- that turns out to be false for the shell case, because the shell test cases show an essentially empty shell (i.e. it has no transitive dependencies of its own).  However, we did talk about how a shell should eagerly include all of its own transitive dependencies directly since it will need them and it is loaded with everything else, so now the shell reliably gobbles up all of its own transitive dependencies now, because its composition includes the new `generateEagerMergeStrategy`.

I should point out that by revisiting the way these strategy generators were built, it actually revealed a previously undiscovered _error_ in the `generateSharedDepsMergeStrategy`, which you can see in the fix to the `shards_test.ts` around line 75 in the test called 'with 3 entrypoints, all deps are in their places'.

The previous strategy was incorrectly bundling `dep1` into the common bundle.  Common, being an entrypoint, should not have been pulled into a bundle which did not represent its own transitive dependencies, but that's what was happening.  Fixing the strategies required updating that test to demonstrate how `shared_bundle_1.html` is created as a result.  The remainder of changes to that `shards_test.ts` file are cosmetic.

There are a lot of new tests in the `bundle-manifest_test.ts` doc but I also simplified the tests stylistically so the existing tests there don't prove anything new, they just are easier to read (i think).

(Note I included some `//` comments with no text afterwards in a couple of lines just to force the formatter to splay the array contents across multiple lines.)

### DIFF
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -32,8 +32,6 @@ export type BundleUrlMapper = (bundles: Bundle[]) => Map<UrlString, Bundle>;
  */
 export type TransitiveDependenciesMap = Map<UrlString, Set<UrlString>>;
 
-
-
 /**
  * A bundle is a grouping of files which serve the need of one or more
  * entrypoint files.
@@ -150,6 +148,51 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
 }
 
 /**
+ * Chains multiple bundle strategy functions together so the output of one
+ * becomes the input of the next and so-on.
+ */
+export function composeStrategies(strategies: BundleStrategy[]) {
+  return strategies.reduce((s1, s2) => (b) => s2(s1(b)));
+}
+
+/**
+ * Generates a strategy function which finds all non-entrypoint bundles which
+ * are dependencies of the given entrypoint and merges them into that
+ * entrypoint's bundle.
+ */
+export function generateEagerMergeStrategy(entrypoint: UrlString):
+    BundleStrategy {
+  return generateMatchMergeStrategy(
+      entrypoint,
+      b => b.entrypoints.has(entrypoint) && !getBundleEntrypoint(b));
+}
+
+/**
+ * Generates a strategy function which finds all bundles matching the predicate
+ * function and merges them into the bundle containing the target file.
+ */
+export function generateMatchMergeStrategy(
+    target: UrlString,
+    predicate: (matchBundle: Bundle, targetBundle: Bundle) =>
+        boolean): BundleStrategy {
+  return (bundles: Bundle[]): Bundle[] => {
+    let targetBundle = bundles.find(bundle => bundle.files.has(target));
+    if (!targetBundle) {
+      throw new Error(`No bundle found containing file ${target}`);
+    }
+    const newBundles = Array.from(bundles);
+    const bundlesToMerge = newBundles.filter(
+        matchBundle => matchBundle === targetBundle ||
+            predicate(matchBundle, targetBundle!));
+    for (const bundle of bundlesToMerge) {
+      newBundles.splice(newBundles.indexOf(bundle), 1);
+    }
+    newBundles.push(mergeBundles(bundlesToMerge));
+    return newBundles;
+  };
+}
+
+/**
  * Generates a strategy function to merge all bundles where the dependencies
  * for a bundle are shared by at least 2 entrypoints (default; set
  * `minEntrypoints` to change threshold).
@@ -166,34 +209,10 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
   if (minEntrypoints === undefined) {
     minEntrypoints = 2;
   }
-  return (bundles: Bundle[]): Bundle[] => {
-    const newBundles: Bundle[] = [];
-    const sharedBundles: Bundle[] = [];
-    const allEntrypoints = new Set<UrlString>();
-    for (const bundle of bundles) {
-      bundle.entrypoints.forEach(
-          (entrypoint) => allEntrypoints.add(entrypoint));
-      if (bundle.entrypoints.size >= minEntrypoints &&
-          !getBundleEntrypoint(bundle)) {
-        sharedBundles.push(bundle);
-      } else {
-        newBundles.push(
-            new Bundle(new Set(bundle.entrypoints), new Set(bundle.files)));
-      }
-    }
-    for (let i = 0; i < newBundles.length; i++) {
-      const bundle = newBundles[i];
-      if (setEquals(bundle.entrypoints, allEntrypoints)) {
-        newBundles.splice(i, 1)[0];
-        sharedBundles.push(bundle);
-        break;
-      }
-    }
-    if (sharedBundles.length > 0) {
-      newBundles.push(mergeBundles(sharedBundles));
-    }
-    return newBundles;
-  };
+  return (bundles: Bundle[]): Bundle[] => mergeMatchingBundles(
+             bundles,
+             bundle => bundle.entrypoints.size >= minEntrypoints &&
+                 !getBundleEntrypoint(bundle));
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/bundle-manifest.ts
```diff
@@ -150,6 +148,51 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
 }
 
 /**
+ * Chains multiple bundle strategy functions together so the output of one
+ * becomes the input of the next and so-on.
+ */
+export function composeStrategies(strategies: BundleStrategy[]) {
+  return strategies.reduce((s1, s2) => (b) => s2(s1(b)));
```
justinfagnani: such higher-order. much laziness.

src/bundle-manifest.ts
```diff
@@ -150,6 +148,51 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
...
+/**
+ * Generates a strategy function which finds all non-entrypoint bundles which
+ * are dependencies of the given entrypoint and merges them into that
+ * entrypoint's bundle.
+ */
+export function generateEagerMergeStrategy(entrypoint: UrlString):
+    BundleStrategy {
+  return generateMatchMergeStrategy(
+      entrypoint,
+      b => b.entrypoints.has(entrypoint) && !getBundleEntrypoint(b));
```
justinfagnani: nit: parens around arrow params, even when there's only 1

src/bundle-manifest.ts
```diff
@@ -150,6 +148,51 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
...
+}
+
+/**
+ * Generates a strategy function which finds all bundles matching the predicate
+ * function and merges them into the bundle containing the target file.
+ */
+export function generateMatchMergeStrategy(
+    target: UrlString,
+    predicate: (matchBundle: Bundle, targetBundle: Bundle) =>
```
justinfagnani: this signature has two parameters, but the call at L165 only has one argument. Is the second param optional?

src/bundle-manifest.ts
```diff
@@ -150,6 +148,51 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
...
+        boolean): BundleStrategy {
+  return (bundles: Bundle[]): Bundle[] => {
+    let targetBundle = bundles.find(bundle => bundle.files.has(target));
+    if (!targetBundle) {
+      throw new Error(`No bundle found containing file ${target}`);
+    }
+    const newBundles = Array.from(bundles);
+    const bundlesToMerge = newBundles.filter(
+        matchBundle => matchBundle === targetBundle ||
+            predicate(matchBundle, targetBundle!));
```
justinfagnani: I'm surprised you need the `!` operator here because of the check on L183. Maybe change that to an explicit null check?

src/bundle-manifest.ts
justinfagnani: `predicate` could use some docs I think, either in the doc comment for `generateMatchMergeStrategy`, or lift the type out and do it there. Mainly, what are `matchBundle` and `targetBundle`?

src/bundle-manifest.ts
```diff
@@ -150,6 +148,51 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
...
+    for (const bundle of bundlesToMerge) {
+      newBundles.splice(newBundles.indexOf(bundle), 1);
```
justinfagnani: Seems like a lot of these operations would be easier if bundles were a Set

src/bundle-manifest.ts
```diff
@@ -166,54 +209,28 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
...
-      }
-    }
-    if (sharedBundles.length > 0) {
-      newBundles.push(mergeBundles(sharedBundles));
-    }
-    return newBundles;
-  };
+  return (bundles: Bundle[]): Bundle[] => mergeMatchingBundles(
+             bundles,
+             bundle => bundle.entrypoints.size >= minEntrypoints &&
```
justinfagnani: nit: parens around param

src/bundle-manifest.ts
```diff
@@ -166,54 +209,28 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
...
+    shell: UrlString, minEntrypoints?: number): BundleStrategy {
+  if (minEntrypoints === undefined) {
+    minEntrypoints = 2;
+  }
+  return composeStrategies([
+    generateEagerMergeStrategy(shell),
+    generateSharedDepsMergeStrategy(minEntrypoints),
+    generateMatchMergeStrategy(
+        shell,
+        b => b.entrypoints.size >= minEntrypoints && !getBundleEntrypoint(b))
```
justinfagnani: nit: parens around param

src/bundle-manifest.ts
```diff
@@ -166,54 +209,28 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
...
-    if (shellBundle && sharedBundle && shellBundle !== sharedBundle) {
-      newBundles.splice(newBundles.indexOf(shellBundle), 1);
-      newBundles.splice(newBundles.indexOf(sharedBundle), 1);
-      newBundles.push(mergeBundles([shellBundle, sharedBundle]));
-    }
-    return newBundles;
-  };
+    shell: UrlString, minEntrypoints?: number): BundleStrategy {
+  if (minEntrypoints === undefined) {
+    minEntrypoints = 2;
```
justinfagnani: super nit: I think `minEntrypoints = minEntrypoints || 2` is fine even though it's not 100% correct in all cases.

src/bundle-manifest.ts
```diff
@@ -166,54 +209,28 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
...
+  }
+  return composeStrategies([
+    generateEagerMergeStrategy(shell),
+    generateSharedDepsMergeStrategy(minEntrypoints),
+    generateMatchMergeStrategy(
```
justinfagnani: Since we've already pulled all shell dependencies into the shell bundle with the eager strategy, is this even necessary anymore?

justinfagnani:
Looks good. Only real question is whether the shell strategy needs to use the match strategy.

src/bundle-manifest.ts
usergenic: Commit 9ff1c27640d06adcb2f6998b08913103601e4100 took care of all instances of non-parenthesized anonymous function arguments.

src/bundle-manifest.ts
usergenic: Turns out that none of the uses of the `generateMatchMergeStrategy` wound up needing access to the `targetBundle` argument in the predicate (i.e. the bundle which contains the file referenced by the `target` argument of strategy generator.  I had thought they might, but nope.  So that's gone and now the predicate's remaining argument is unambiguous.

src/bundle-manifest.ts
usergenic: I can't think for the life of me why 0 would be a good choice for minEntrypoints but its falsey which would `||` to 2 undesireably.

src/bundle-manifest.ts
usergenic: For some reason the `if (!targetBundle)` with a throw in it didn't seem to satisfy the type checker and it thought the type was still `Bundle | undefined`, so I `!`'d it.

moot point now, since I don't pass it into that function anymore.

src/bundle-manifest.ts
usergenic: Hmm.  Perhaps.  May play with that another time.

usergenic: I don't want to redefine strategy interface from Array to Set today and I feel like that might be the way to go.

src/bundle-manifest.ts
usergenic: done

src/bundle-manifest.ts
usergenic: Per discussion earlier, it turns out that the generateSharedDepsMergeStrategy is actually the redundant one.  We need the generateMatchMergeStrategy because that's what the expectation is for shop-- there's all these shared dependencies which should be inlined into the shell but none of them are transitive dependencies of shell, and the match merge is the one that does that.

usergenic: I actually decided to update generateMatchMergeStrategy now to simply merge together everything matching the predicate and its the caller that determines how to use that now.  This means that the generateSharedDepsMergeStrategy and generateEagerMergeStrategy and generateShellMergeStrategy all use it now.

usergenic:
Okay I'm done whittling it down.  It's fairly whittled.  Time for bed.  Whenever you're free next etc, @justinfagnani PTAL :)

src/bundle-manifest.ts
justinfagnani: Speaking of, I guess this could use some validation. `0` is probably bad. `-1` is right out.

PR #428, opened 2017-02-12
Update to latest analyzer.

rictic:

### DIFF
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -188,11 +188,10 @@ export function generateMatchMergeStrategy(predicate: (b: Bundle) => boolean):
  * dependencies which are shared by at least 2 entrypoints:
  *   `[a]->[a,b], [c]->[c,e], [a,c,f,g]->[d,f,g,h]`
  */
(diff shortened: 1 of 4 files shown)


src/bundle-manifest.ts
```diff
@@ -205,10 +204,9 @@ export function generateSharedDepsMergeStrategy(minEntrypoints?: number):
  * bundle for an application shell.
  */
 export function generateShellMergeStrategy(
-    shell: UrlString, minEntrypoints?: number): BundleStrategy {
-  if (minEntrypoints === undefined) {
-    minEntrypoints = 2;
-  }
+    shell: UrlString, maybeMinEntrypoints?: number): BundleStrategy {
+  const minEntrypoints =
+      maybeMinEntrypoints === undefined ? 2 : maybeMinEntrypoints;
```
rictic: It seems that in 2.2.0 tsc lost some ability to do local type narrowing. My guess would be a bug involving `arguments`, as this code relied upon being able to determine that `minEntrypoints` wasn't assigned to.

usergenic: Steve McCroskey

Looks like I picked the wrong time to start reassigning arguments for defaults.

PR #431, opened 2017-02-15
When bundling documents which use the `<base>` tag, emulate it and remove it.

usergenic:
 - Handle `<base href="...">` values correctly when resolving imports.
 - Apply `<base target="...">` to links and forms in the same document as appropriate.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -191,7 +191,6 @@ export class Bundler {
 
     this._appendHtmlImportsForBundle(document, docBundle);
     importUtils.rebaseDocument(docUrl, document);
-    importUtils.rewriteImportedUrls(document, docUrl, docUrl);
 
     await this._inlineHtmlImports(docUrl, document, docBundle, bundleManifest);
 
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -188,7 +188,9 @@ export class Bundler {
       bundleImports?: Set<string>): Promise<ASTNode> {
     const docUrl = docBundle.url;
     const document = await this._prepareBundleDocument(docBundle);
+
     this._appendHtmlImportsForBundle(document, docBundle);
+    importUtils.rebaseDocument(docUrl, document);
     importUtils.rewriteImportedUrls(document, docUrl, docUrl);
 
     await this._inlineHtmlImports(docUrl, document, docBundle, bundleManifest);
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -190,7 +190,7 @@ export class Bundler {
     const document = await this._prepareBundleDocument(docBundle);
 
     this._appendHtmlImportsForBundle(document, docBundle);
-    importUtils.rebaseDocument(docUrl, document);
+    importUtils.debaseDocument(docUrl, document);
 
     await this._inlineHtmlImports(docUrl, document, docBundle, bundleManifest);
 
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -190,7 +190,7 @@ export class Bundler {
     const document = await this._prepareBundleDocument(docBundle);
 
     this._appendHtmlImportsForBundle(document, docBundle);
-    importUtils.debaseDocument(docUrl, document);
+    importUtils.rewriteDocumentToEmulateBaseTag(docUrl, document);
 
     await this._inlineHtmlImports(docUrl, document, docBundle, bundleManifest);
 
--- a/src/deps-index.ts
+++ b/src/deps-index.ts
@@ -35,7 +35,7 @@ async function _getTransitiveDependencies(
     url: UrlString, entrypoints: UrlString[], analyzer: Analyzer):
     Promise<DependencyMapEntry> {
       const document = await analyzer.analyze(url);
-      const documentBase = document.parsedDocument.baseUrl;
+      const baseUrl = document.parsedDocument.baseUrl;
       const imports = document.getByKind(
           'import', {externalPackages: true, imported: true});
       const eagerImports = new Set<UrlString>();
@@ -51,7 +51,7 @@ async function _getTransitiveDependencies(
           throw err;
         }
         const resolvedHtmlImportUrl = urlLib.resolve(
-            documentBase, urlUtils.relativeUrl(documentBase, htmlImport.url));
+            baseUrl, urlUtils.relativeUrl(baseUrl, htmlImport.url));
         switch (htmlImport.type) {
           case 'html-import':
             eagerImports.add(resolvedHtmlImportUrl);
--- a/src/deps-index.ts
+++ b/src/deps-index.ts
@@ -13,9 +13,6 @@
  */
 import {AssertionError} from 'assert';
 import {Analyzer} from 'polymer-analyzer';
-import * as urlLib from 'url';
-
-import * as urlUtils from './url-utils';
 import {UrlString} from './url-utils';
 
 export interface DepsIndex {
@@ -35,7 +32,6 @@ async function _getTransitiveDependencies(
     url: UrlString, entrypoints: UrlString[], analyzer: Analyzer):
     Promise<DependencyMapEntry> {
       const document = await analyzer.analyze(url);
-      const baseUrl = document.parsedDocument.baseUrl;
       const imports = document.getByKind(
           'import', {externalPackages: true, imported: true});
       const eagerImports = new Set<UrlString>();
@@ -50,14 +46,13 @@ async function _getTransitiveDependencies(
           }
           throw err;
         }
-        const resolvedHtmlImportUrl = urlLib.resolve(
-            baseUrl, urlUtils.relativeUrl(baseUrl, htmlImport.url));
+
         switch (htmlImport.type) {
           case 'html-import':
-            eagerImports.add(resolvedHtmlImportUrl);
+            eagerImports.add(htmlImport.url);
             break;
           case 'lazy-html-import':
-            lazyImports.add(resolvedHtmlImportUrl);
+            lazyImports.add(htmlImport.url);
             break;
         }
       }
--- a/src/deps-index.ts
+++ b/src/deps-index.ts
@@ -13,12 +13,14 @@
  */
 import {AssertionError} from 'assert';
 import {Analyzer} from 'polymer-analyzer';
+import * as urlLib from 'url';
 
+import * as urlUtils from './url-utils';
 import {UrlString} from './url-utils';
 
 export interface DepsIndex {
   // An index of entrypoint -> html dependencies
-  entrypointToDeps: Map<string, Set<string>>;
+  entrypointToDeps: Map<UrlString, Set<UrlString>>;
 }
 
 type DependencyMapEntry = {
@@ -33,11 +35,12 @@ async function _getTransitiveDependencies(
     url: UrlString, entrypoints: UrlString[], analyzer: Analyzer):
     Promise<DependencyMapEntry> {
       const document = await analyzer.analyze(url);
+      const documentBase = document.parsedDocument.baseUrl;
       const imports = document.getByKind(
           'import', {externalPackages: true, imported: true});
       const eagerImports = new Set<UrlString>();
       const lazyImports = new Set<UrlString>();
-      for (let htmlImport of imports) {
+      for (const htmlImport of imports) {
         try {
           console.assert(
               htmlImport.url, 'htmlImport: %s has no url', htmlImport);
@@ -47,12 +50,14 @@ async function _getTransitiveDependencies(
           }
           throw err;
         }
+        const resolvedHtmlImportUrl = urlLib.resolve(
+            documentBase, urlUtils.relativeUrl(documentBase, htmlImport.url));
         switch (htmlImport.type) {
           case 'html-import':
-            eagerImports.add(htmlImport.url);
+            eagerImports.add(resolvedHtmlImportUrl);
             break;
           case 'lazy-html-import':
-            lazyImports.add(htmlImport.url);
+            lazyImports.add(resolvedHtmlImportUrl);
             break;
         }
       }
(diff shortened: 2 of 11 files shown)


src/test/bundler_test.ts
```diff
@@ -94,13 +94,13 @@ suite('Bundler', () => {
...
 
-  test.skip('Handle <base> tag', () => {
+  test('Handle <base> tag', () => {
     const span = preds.AND(
         preds.hasTagName('span'), preds.hasAttrValue('href', 'imports/hello'));
     const a = preds.AND(
         preds.hasTagName('a'),
         preds.hasAttrValue('href', 'imports/sub-base/sub-base.html'));
-    return bundle('html/base.html').then((doc) => {
+    return bundle('test/html/base.html').then((doc) => {
```
rictic: Recommend using async await

src/matchers.ts
```diff
@@ -84,6 +84,9 @@ export const hiddenDiv: Matcher = predicates.AND(
     predicates.hasAttr('hidden'),
     predicates.hasAttr('by-polymer-bundler'));
 export const inHiddenDiv: Matcher = predicates.parentMatches(hiddenDiv);
+export const baseTargetAppliesTo: Matcher = predicates.AND(
```
rictic: As this is only used one place, recommend placing it in that file. More legible, easier to evaluate that it does what's expected.

src/import-utils.ts
```diff
@@ -11,25 +11,48 @@
...
 
 // TODO(usergenic): Revisit the organization of this module and *consider*
 // building a class to encapsulate the common document details like docUrl and
 // docBundle and global notions like manifest etc.
 
+/*
+ * Given an import document with a base tag, transform all of its URLs and set
+ * link and form target attributes and remove the base tag.
+ */
+export async function debaseDocument(docUrl: UrlString, importDoc: ASTNode) {
```
rictic: Recommend removing the async here. Callers don't seem to be awaiting on it, and it doesn't have any awaits itself

src/bundler.ts
```diff
@@ -188,8 +188,9 @@ export class Bundler {
       bundleImports?: Set<string>): Promise<ASTNode> {
     const docUrl = docBundle.url;
     const document = await this._prepareBundleDocument(docBundle);
+
     this._appendHtmlImportsForBundle(document, docBundle);
-    importUtils.rewriteImportedUrls(document, docUrl, docUrl);
+    importUtils.debaseDocument(docUrl, document);
```
justinfagnani: is there a better name than `debaseDocument`? `applyBaseUrl`?

src/deps-index.ts
```diff
@@ -13,12 +13,14 @@
...
 import {Analyzer} from 'polymer-analyzer';
+import * as urlLib from 'url';
 
+import * as urlUtils from './url-utils';
 import {UrlString} from './url-utils';
 
 export interface DepsIndex {
   // An index of entrypoint -> html dependencies
-  entrypointToDeps: Map<string, Set<string>>;
+  entrypointToDeps: Map<UrlString, Set<UrlString>>;
```
justinfagnani: I'm not sure how much `UrlString` is buying us, since it's a structural interface. We should reconsider this, but not in this PR.

src/deps-index.ts
```diff
@@ -33,11 +35,12 @@ async function _getTransitiveDependencies(
     url: UrlString, entrypoints: UrlString[], analyzer: Analyzer):
     Promise<DependencyMapEntry> {
       const document = await analyzer.analyze(url);
+      const documentBase = document.parsedDocument.baseUrl;
```
justinfagnani: Given the very general meaning of `base`, I'd suggest going long with `documentBaseUrl`, or maybe just `baseUrl`.

src/url-utils.ts
```diff
@@ -30,6 +30,19 @@ const sharedRelativeUrlProperties =
 export type UrlString = string;
 
 /**
+ * Returns a URL with the basename removed from the pathname.  Strips the
+ * search off of the URL as well, since it will not apply.
+ */
+export function debaseUrl(href: UrlString): UrlString {
```
rictic: debase doesn't seem like the right name here. 

It's used in few enough places that it seems worth giving it a longer, more descriptive name. removeSearchAndFilename?

src/url-utils.ts
```diff
@@ -30,6 +30,19 @@ const sharedRelativeUrlProperties =
...
+  const u = url.parse(href);
+  if (typeof u.pathname === 'string') {
+    u.pathname = path.dirname(u.pathname + '_').replace(/_$/, '') + '/';
+  }
+  u.search = undefined;
```
rictic: Should hashes be removed as well?

src/url-utils.ts
```diff
@@ -30,6 +30,19 @@ const sharedRelativeUrlProperties =
 export type UrlString = string;
 
 /**
+ * Returns a URL with the basename removed from the pathname.  Strips the
+ * search off of the URL as well, since it will not apply.
+ */
+export function debaseUrl(href: UrlString): UrlString {
+  const u = url.parse(href);
+  if (typeof u.pathname === 'string') {
+    u.pathname = path.dirname(u.pathname + '_').replace(/_$/, '') + '/';
```
rictic: What's with adding and removing the underscore here?

src/url-utils.ts
```diff
@@ -30,6 +30,19 @@ const sharedRelativeUrlProperties =
 export type UrlString = string;
 
 /**
+ * Returns a URL with the basename removed from the pathname.  Strips the
+ * search off of the URL as well, since it will not apply.
+ */
+export function debaseUrl(href: UrlString): UrlString {
+  const u = url.parse(href);
+  if (typeof u.pathname === 'string') {
```
rictic: What other types might pathname be? If it's just null/undefined, recommend instead 

```typescript
if (u !== null) {
  ...
}
```

I'd prefer that style because it documents for the reader that u.pathname is either a string or null/undefined (otherwise typescript would warn).

src/deps-index.ts
```diff
@@ -47,12 +50,14 @@ async function _getTransitiveDependencies(
           }
           throw err;
         }
+        const resolvedHtmlImportUrl = urlLib.resolve(
+            documentBase, urlUtils.relativeUrl(documentBase, htmlImport.url));
```
justinfagnani: why do we need both `resolve` and `relativeUrl` calls, with `documentBase` appearing twice? Can't `resolve` alone do it?

src/import-utils.ts
```diff
@@ -11,25 +11,48 @@
...
+  const baseTag = dom5.query(importDoc, matchers.base);
+  // If there's no base tag, there's nothing to do.
+  if (!baseTag) {
+    return;
+  }
+  for (const baseTag of dom5.queryAll(importDoc, matchers.base)) {
+    astUtils.removeElementAndNewline(baseTag);
```
justinfagnani: I wonder if we should warn if there's more than one base tag

src/test/import-utils_test.ts
```diff
@@ -62,132 +65,138 @@ suite('import-utils', () => {
...
+        <template>
+        <style>:host { background-image: url("my-element/background.svg"); }</style>
+        </template>
+        </dom-module>
+        <script>Polymer({is: "my-element"})</script></body></html>
+      `;
 
       const ast = parse5.parse(htmlBase);
-      // pathRewriter.acid(ast, inputPath);
+      importUtils.debaseDocument(importDocPath, ast);
```
justinfagnani: Can debaseDocument be tested in isolation in it's own suite?

src/import-utils.ts
```diff
@@ -94,6 +117,7 @@ export async function inlineHtmlImport(
   }
 
   const importDoc = parse5.parseFragment(importSource);
+  debaseDocument(resolvedImportUrl, importDoc);
```
justinfagnani: Why is debaseDocument a pass on the document, rather than just a part of `rewriteImportedUrls`?

src/test/bundler_test.ts
usergenic: I converted all 41 tests in `bundler_test.ts` to `async await`.  I assume these were all `.then` because ported over from pre-typescript vulcanize.

src/matchers.ts
usergenic: Moved.

src/import-utils.ts
usergenic: Good point.  Deasync'd.

src/url-utils.ts
usergenic: Renamed to stripUrlFileSearchAndHash

src/url-utils.ts
usergenic: Fair.  Weird issue-- typescript says it should be string | undefined, but the url.parse returns an object with nulls on it and doesn't let me assign nulls.  Anyways I corrected it to do a (u.pathname != null) which does the right thing and doesn't string match null, so it should be good enough now I think.

src/url-utils.ts
usergenic: I'll add a comment.  The reason for this is `path.dirname('/a/b/') incorrectly returns `/a`.  By suffixing the string with `_` I ensure that `/a/b/` is treated as `/a/b/_` and returns `/a/b`.  Also, `/a/b` is treated as `/a/b_` and therefore returns `/a` correctly.

The removing of the underscore was redundant as there should be no circumstance it remains.  That was a bug!  Thanks for bringing it up.  I have another code point where I do remove the underscore for a similar trick but that is not a bug.  This was.  Fixed and commented!

src/url-utils.ts
usergenic: Now they are.  Good point.

src/deps-index.ts
usergenic: Buys nothing with type checking.  I guess I liked the idea of them documenting the intent/content of the strings, but I haven't found a way to leverage usefully.  I'm open to trading for `string` at some point.

src/test/url-utils_test.ts
```diff
@@ -26,23 +26,26 @@ const assert = chai.assert;
 
 suite('URL Utils', () => {
```
rictic: Add a test with a hash?

src/bundler.ts
usergenic: If its okay, I've decided to rename it `rewriteDocumentToEmulateBaseTag`.  It actually applies two effects.  One is that it adjusts the urls accordingly.  Two is that it applies the "target" attribute of the base tag to links and forms which don't have a target.  I also added a regression test to ensure isolation of the effects of target attribute application.

src/deps-index.ts
usergenic: Renamed to baseUrl

src/test/url-utils_test.ts
usergenic: Done.

src/deps-index.ts
usergenic: So this is embarrassing.  Turns out we don't need to resolve the urls at all now, because the analyzer does the work.  This code is from when bundler was trying to do the base url resolution on its own.  Removed dependency on `url` lib and `url-utils`'s `relativeUrl`.

src/import-utils.ts
usergenic: Well, we don't need to call `rewriteImportedUrls` on our bundle entrypoint doc, but we *do* need to call it when we `inlineHtmlImport`.  However, we need to call `rewriteDocumentToEmulateBaseTag` (formerly `debaseDocument`) on all documents.  That's why they're split up.

usergenic: Thinking about it further, since the purpose of `rewriteImportedUrls` is to rewrite urls in a document because the base they were build on has shifted, the term "Imported" is misleading, since we actually have two contexts where we manipulate the base of a document:

1. We encounter a base tag and we have to remove it and emulate it because we're merging documents together.
2. We inline a document and have to modify the contents of it to be as if it has a new base url.

So, for clarity, I've renamed a lot of things:
```
export function debaseDocument -> rewriteDocumentToEmulateBaseTag
export function rewriteImportedRelPath -> rewriteHrefBaseUrl
export function rewriteImportedUrls -> rewriteDocumentBaseUrl
function _rewriteImportedElementAttrUrls ->  _rewriteElementAttrsBaseUrl
function _rewriteImportedStyleTextUrls -> _rewriteCssTextBaseUrl
function _rewriteImportedStyleUrls -> _rewriteStyleTagsBaseUrl
function _setImportedDomModuleAssetpaths -> _setDomModuleAssetpaths
```

Also the parameters to all the functions have been renamed:
```
importDoc -> ast
importDocUrl -> oldBaseUrl
mainDocUrl -> newBaseUrl
```

I've also normalized the argument order of the functions so that the target (text or ast node) comes first, followed by oldBaseUrl and newBaseUrl.

src/test/import-utils_test.ts
usergenic: `Document <base> tag emulation` is its own suite now with focus only on the `rewriteDocumentToEmulateBaseTag` function effects now.

usergenic:
@justinfagnani all change requests responded to and all refactoring committed.  PTAL

src/import-utils.ts
```diff
@@ -176,40 +175,81 @@ export async function inlineStylesheet(
...
+  _rewriteStyleTagsBaseUrl(ast, oldBaseUrl, newBaseUrl);
+  _setDomModuleAssetpaths(ast, oldBaseUrl, newBaseUrl);
+}
+
+/**
+ * Given a string of CSS, return a version where all occurrences of urls,
+ * have been rewritten based on the relationship of the old base url to the
+ * new base url.
+ */
+function _rewriteCssTextBaseUrl(
```
justinfagnani: nit: non-exported functions don't need to be named with an underscore

usergenic: Noted and fixed!

Issue #427, opened 2017-02-09
SSI tags are being stripped as comments


pbcomm:
Vulcanize/Bundler strips SSI (Server Side Include) tags.

Ex: 
`<!--# if expr="${SERVER_VAR} = something" -->`
`<!--#echo var='someVarName' -->`

usergenic:
This is fixed in master.  Will go out in next release 2.0.0-pre.6

PR #432, opened 2017-02-16
Rewrite stuff in templates too

usergenic:
- The url rewriter was not descending into `<template>` tags to do its business.  That was bad #430 but now its good.
 - [x] CHANGELOG.md has been updated

PR #433, opened 2017-02-17
Dedup license comments

usergenic:
 - There were cases in my builds I saw tons of duplicated licenses.
 - I wanted to fix that.
 - It was due to whitespace issues mostly.
 - I had to monkey with the strip comments code because #427 and have a separate PR for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Handle `<base href="...">` values correctly when inlining imports.
 - Apply `<base target="...">` to links and forms in the same document as
   appropriate.
+- License comments are deduplicated properly now.
 
 ## 2.0.0-pre.5 - 2017-02-14
 - Handle base tags when resolving the dependency graph of imports.
--- a/src/ast-utils.ts
+++ b/src/ast-utils.ts
@@ -109,16 +109,21 @@ export function siblingsAfter(node: ASTNode): ASTNode[] {
  * deduplicate them and prepend them in document's head.
  */
 export function stripComments(document: ASTNode) {
-  // Use of a Map keyed by comment text enables deduplication.
-  const comments: Map<string, ASTNode> = new Map();
-  dom5.nodeWalkAll(document, dom5.isCommentNode).forEach((comment: ASTNode) => {
-    comments.set(comment.data || '', comment);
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Handle `<base href="...">` values correctly when inlining imports.
 - Apply `<base target="...">` to links and forms in the same document as
   appropriate.
+- Fixed a bug where element attributes inside `<template>` tags were not
```
rictic: Does this PR also include https://github.com/Polymer/polymer-bundler/pull/432?

src/ast-utils.ts
```diff
@@ -109,16 +109,21 @@ export function siblingsAfter(node: ASTNode): ASTNode[] {
...
-  dom5.nodeWalkAll(document, dom5.isCommentNode).forEach((comment: ASTNode) => {
-    comments.set(comment.data || '', comment);
-    removeElementAndNewline(comment);
-  });
-  const head = dom5.query(document, matchers.head);
-  for (const comment of comments.values()) {
-    if (isLicenseComment(comment)) {
-      prepend(head || document, comment);
+  const uniqueLicenseTexts = new Set<string>();
+  const licenseComments: ASTNode[] = [];
```
rictic: No need to keep an array too, Sets are ordered in javascript.

CHANGELOG.md
rictic: Prefer either doing parallel PRs or using github's base branch feature to chain PRs off of one another. Otherwise it's tricky to tell in which PR I should leave a particular comment.

src/ast-utils.ts
usergenic: The array contains the actual nodes though.  I use the set just to deduplicate the text in a whitespace agnostic form.

CHANGELOG.md
usergenic: You're right.  I should have set the base on the PR differently.  I think I was hoping to get the other one in master first and then the extra stuff would disappear before this was looked at.  

Now I've rebased the one I cut this from so I've rebased this one too.

src/ast-utils.ts
rictic: I see, makes sense.

PR #439, opened 2017-02-25
Going all in on polymer analyzer

### LINKED ISSUE #436: External scripts are stripped out when bundling
When using `google-chart` element with the `polymer-build` 0.8.1 and `polymer-bundler` 2.0.0-pre.6, [this line](https://github.com/GoogleWebComponents/google-chart/blob/master/charts-loader.html#L4) is stripped out of the bundle, which results in the following error:
```
Uncaught ReferenceError: google is not defined
    at HTMLElement.<anonymous> (sandbox-demo-first-sub-page.js:1)
    at i.complete (sandbox-app.js:2)
    at boundComplete (sandbox-app.js:2)
```

I updated my [sandbox](https://github.com/web-padawan/polymer-sandbox) to reproduce this issue, see it live [here](https://sandbox-e0d4b.firebaseapp.com/demo/first/) and open console to see that error.

AFAIK there used to be a `NoopResolver` for such scripts when using `vulcanize` and `hydrolysis`. Looks like this logic is broken for some reasons.

usergenic:
- Uses `polymer-analyzer` directly to obtain import document sources and
   ASTs.  This enforces agreement with the Analyzer as to which documents
   should be inlined/bundled.  (Fixes #436)
 - Fixes issue where protocol-less URLs (those that start with `//`) were
   treated as absolute paths.
 - Renamed/refactored to draw a clearer distinction between ASTs and
   documents.

### DIFF
--- a/src/ast-utils.ts
+++ b/src/ast-utils.ts
@@ -13,17 +13,9 @@
  */
 
 import * as dom5 from 'dom5';
-import {ASTNode, parse, serialize} from 'parse5';
+import {ASTNode} from 'parse5';
 
 import * as matchers from './matchers';
-
-/**
- * Creates a clone of the AST.  The clone is hygenic, by reparsing the
- * serialized form of the given AST.
- */
-export function cloneTree(ast: ASTNode): ASTNode {
-  return parse(serialize(ast));
-}
 
 /**
  * Move the `node` to be the immediate sibling after the `target` node.
--- a/src/ast-utils.ts
+++ b/src/ast-utils.ts
@@ -13,9 +13,17 @@
  */
 
 import * as dom5 from 'dom5';
-import {ASTNode} from 'parse5';
+import {ASTNode, parse, serialize} from 'parse5';
 
 import * as matchers from './matchers';
+
+/**
+ * Creates a clone of the AST.  The clone is hygenic, by reparsing the
+ * serialized form of the given AST.
+ */
+export function cloneTree(ast: ASTNode): ASTNode {
+  return parse(serialize(ast));
+}
 
 /**
  * Move the `node` to be the immediate sibling after the `target` node.
--- a/src/import-utils.ts
+++ b/src/import-utils.ts
@@ -117,11 +117,11 @@ export async function inlineHtmlImport(
  * into the script tag content and removes the src attribute.
  */
 export async function inlineScript(document: Document, scriptTag: ASTNode) {
-  const rawUrl = dom5.getAttribute(scriptTag, 'src')!;
-  const resolvedUrl = urlLib.resolve(document.url, rawUrl);
+  const rawImportUrl = dom5.getAttribute(scriptTag, 'src')!;
+  const resolvedImportUrl = urlLib.resolve(document.url, rawImportUrl);
   const scriptImport = findInSet(
       document.getByKind('html-script', {imported: true}),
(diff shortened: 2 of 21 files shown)


usergenic:
A lot of code simply renames `document` to `ast` in places that I'm dealing with `ASTNode` instead of `Document` objects as a way to draw a clear distinction and not use the terms interchangeably.  This is because the rest of the changes involve passing `Document` objects into the import utility functions so they can use `getByKind` to ask for imported document contents or ASTs.

src/import-utils.ts
```diff
@@ -73,97 +73,88 @@ export async function inlineHtmlImport(
...
-  try {
-    script = await loader(resolvedUrl);
-  } catch (e) {
-    // If a script doesn't load, skip inlining.
-    // TODO(usergenic): Add plylog logging for load error.
-    console.warn(`Unable to load script at ${resolvedUrl}: ${e.message}`);
-  }
-
-  if (script === undefined) {
+export async function inlineScript(document: Document, scriptTag: ASTNode) {
```
justinfagnani: This seems like it'll work as long as the document already had the imported URL in its transitive imports. Otherwise, getByKind won't return anything. You might need to pass in a collection of documents?

I'd make sure you have a test for this scenario.

src/import-utils.ts
```diff
@@ -73,97 +73,88 @@ export async function inlineHtmlImport(
...
+  const rawImportUrl = dom5.getAttribute(scriptTag, 'src')!;
+  const resolvedImportUrl = urlLib.resolve(document.url, rawImportUrl);
+  const scriptImport = findInSet(
+      document.getByKind('html-script', {imported: true}),
```
justinfagnani: You need `externalPackages: true` also. Add test.

src/ast-utils.ts
```diff
@@ -13,11 +13,19 @@
...
+import {ASTNode, parse, serialize} from 'parse5';
 
 import * as matchers from './matchers';
 
 /**
+ * Creates a clone of the AST.  The clone is hygenic, by reparsing the
+ * serialized form of the given AST.
+ */
+export function cloneTree(ast: ASTNode): ASTNode {
+  return parse(serialize(ast));
```
usergenic: Switched to `clone` and removed `cloneTree` function.

src/import-utils.ts
usergenic: Added tests to show that bundles which include files not present in the source document successfully have those files inlined and that scripts and css linked to from within html imports added to bundles by the strategy are also inlined.

src/import-utils.ts
usergenic: Added a `'html imports from bower_components are inlined'` test

usergenic:
@justinfagnani addressed feedback PTAL

usergenic:
(Note: this PR is not dependent on the new `Analyzer#_fork` method-- that will be in the build PR)

web-padawan:
@usergenic should this fix #436? Please see [my comment](https://github.com/Polymer/polymer-bundler/issues/436#issuecomment-282975732) about external scripts, those are still stripped out with latest `polymer-build` 0.8.2

usergenic:
@web-padawan yes this whole refactor came about as a result of addressing the underlying reasons why #436 was happening.  Thanks for the great bug reports btw!

PR #448, opened 2017-03-06
The stripExcludes option is being completely ignored

ChadKillingsworth:
There is an existing test that is passing incorrectly. The stripExcludes option is never referenced.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -86,6 +86,8 @@ export class Bundler {
     // implicitStrip should be true by default
     this.implicitStrip = !Boolean(opts.noImplicitStrip);
     this.excludes = Array.isArray(opts.excludes) ? opts.excludes : [];
+    this.stripExcludes =
+        Array.isArray(opts.stripExcludes) ? opts.stripExcludes : [];
     this.stripComments = Boolean(opts.stripComments);
     this.enableCssInlining = Boolean(opts.inlineCss);
     this.enableScriptInlining = Boolean(opts.inlineScripts);
@@ -310,6 +312,7 @@ export class Bundler {
           visitedUrls,
           bundle,
           bundleManifest,
+          this.stripExcludes,
           this.sourcemaps);
     }
   }
--- a/src/import-utils.ts
+++ b/src/import-utils.ts
@@ -41,6 +41,7 @@ export async function inlineHtmlImport(
     visitedUrls: Set<UrlString>,
     docBundle: AssignedBundle,
     manifest: BundleManifest,
+    stripExcludes: UrlString[],
     enableSourcemaps: boolean) {
   const rawImportUrl = dom5.getAttribute(linkTag, 'href')!;
   const importUrl = urlLib.resolve(document.url, rawImportUrl);
@@ -56,6 +57,13 @@ export async function inlineHtmlImport(
   // We've never seen this import before, so we'll add it to the set to guard
   // against processing it again in the future.
   visitedUrls.add(resolvedImportUrl);
+
+  // Stripped excluded imports should simply be removed from the DOM
+  if (stripExcludes.find(exclude => resolvedImportUrl.indexOf(exclude) >= 0) !==
+      undefined) {
+    astUtils.removeElementAndNewline(linkTag);
+    return;
+  }
 
   // If we can't find a bundle for the referenced import, record that we've
   // processed it, but don't remove the import link.  Browser will handle it.
@@ -131,6 +139,7 @@ export async function inlineHtmlImport(
         visitedUrls,
         docBundle,
         manifest,
+        stripExcludes,
         enableSourcemaps);
   }
 }
--- a/src/test/bundler_test.ts
+++ b/src/test/bundler_test.ts
@@ -322,6 +322,9 @@ suite('Bundler', () => {
 
     const excludes = ['imports/simple-import.html'];
 
+    const excludedModules = preds.AND(
+        preds.hasTagName('dom-module'), preds.hasAttrValue('id', 'my-element'));
+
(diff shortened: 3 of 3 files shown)


ChadKillingsworth:
Rebased onto master to resolve conflicts with #449

rictic:
Would you mind fixing the test that's passing incorrectly?

Also IIRC @justinfagnani has opinions about stripExcludes, perhaps he should review this

ChadKillingsworth:
Tests fixed.

usergenic:
@ChadKillingsworth thank you for the PR.  We had been considering removing/not-implementing this feature for 2.0, in part because the impetus for the feature in Vulcanize was to support multiple calls into Vulcanize from `polymer-build` or other scripts to produce sharded builds-- but we've implemented the concept of shards/bundles directly into `polymer-bundler` now, so the concept of a `stripExcludes` felt like it might be baggage. 

I'd love to know what your use case is.  Thanks!

ChadKillingsworth:
@usergenic I guess first off you should know that I'd don't use polymer-cli or polymer-build on my projects and really don't have any plans to. My build process is more complicated than most.

Our code base is fully optimized with Closure-Compiler with Advanced Optimizations. In fact, the only thing in our project that is not compatible is Polymer itself. I hear that 2.0 is moving that way, but that doesn't help me in the short term. In addition, all JS dependencies are managed with CommonJS modules until such a time as `<script type="module">` has enough browser support that we can switch to ES6 modules for uncompiled mode.

For these reasons the standard Polymer build tooling has always been challenging.

During build, we strip Polymer itself completely (using stripExcludes). It then gets prepended at the end of the build process. 

The rest of the imports are bundled and piped through cripser to create a single JS file which can the be processed with other standard build tooling (like Closure-compiler).

I've been following the split/rejoin work of polymer-build, but it is not even close to being able to work with our builds.

nverba:
I have a similar situation where we use a more complex build process to meet our needs. We use Polymer to deploy client sites consisting of several bundles. Typically one bundle of core components that we use everywhere, and a client specific bundles with custom UI/user flows. (Our core components are hosted)

So for example,  if we have import a Polymer catalogue element in both bundles, they both pull in a copy of Polymer.html at bundle time and the app will crash on load.  strip-exclude has resolved this for us so far.

ChadKillingsworth:
@usergenic I see that the option was completely removed. I'm now faced with a choice: maintain a fork and private distribution, or develop another post-build tool to strip out the excluded file.

Why remove an option that was used and close to trivial to maintain?

usergenic:
The option was removed because it wasn't implemented and it looked like it was broken -- Didn't want the official release to include an apparently broken/incomplete feature.

I don't think we are necessarily against re-implementing the feature, but the operation to excise the imports ideally should happen prior to the generation of the bundle manifest, in order that strategies which arrange the bundles are not being misled by incorrect dependency graph information.

I could see this working in a few different ways:
1.) adjusting `deps-index.ts` `buildDepsIndex` to consider the array of `stripExcludes` as it traverses the import graph and then the inlining code would also consider them so it knows what to strip.
2.) pre-bundling task strips links _before_ files ever make it to the bundler.
3.) url loaders can strip excluded links when returning the file contents to the analyzer.

Number 3 could be implemented as a class that extends another loader, like FSUrlLoader, and simply performs a transformation of the `load` response before resolving the promise...

@ChadKillingsworth do you use vulcanize/polymer-bundler as a library or the packaged bin vulcanize/polymer-bundler command-line script?

justinfagnani:
To hopefully clarify a little bit, we're viewing Bundler a little differently than Vulcanize, which effects which features it should include. Vulcanize was often the only build tool used in a Polymer project, so it grew a lot of features that weren't specifically related to bundling. Bundler is presumed to be one step among many in a build pipeline, so we'd like it to focus only on bundling.

Modifying the dependency graph in this way - excluding files, removing imports to them, then deciding whether to force exclude transitive dependencies of the file or not - is not really a bundling concern. You can do this modification before the files hit the Bundler and everything will just work.

ChadKillingsworth:
I thought I responded to this ... but I use polymer-bundler as a library.

@justinfagnani You mean just modify the JSON produced by `generateManifest()` before calling `.bundle()`?

justinfagnani:
I mean modify the project files before they hit the bundler.

usergenic:
@ChadKillingsworth - I just released a new version of polymer-bundler and you may be interested in a new property added to Bundle definitions called `stripImports` which does essentially the same job as the `stripExcludes` option, except they are defined and applied per-bundle.

https://github.com/Polymer/polymer-bundler/blob/master/src/bundle-manifest.ts#L49

Right now there's not a shiny wrapper/affordance to define a global exclude, but you can get what you need with something like the following:

```js
import {Bundler, composeStrategies, generateShellMergeStrategy, generateNoBackLinkStrategy} from 'polymer-bundler';

// you can use the strategy to do the job of adding the
// urls to strip with the `generateNoBackLinkStrategy`
// function...
const strategy = composeStrategies(
  generateShellMergeStrategy(2), // or whatever
  generateNoBackLinkStrategy(['url/to/exclude1.html', 'url/to/exclude2.html'])
);
const bundler = new Bundler({strategy: strategy});
const manifest = bundler.generateManifest([...]);
// OR manipulate the bundles in the manifest after the fact:
for (const [url, bundle] of manifest.bundles) {
  bundle.stripImports.push('url/to/exclude1.html');
  bundle.stripImports.push('url/to/exclude2.html');
}
```

Perhaps this will help you get where you need to?

ChadKillingsworth:
@usergenic That should handle my use cases completely! Thanks!

PR #449, opened 2017-03-06
Add sourcemap support

ChadKillingsworth:
Proper sourcemap support is a godsend to debugging and can provide unmangled stack traces from production environments using services like https://sentry.io/  I've been using a modified Vulcanize to generate sourcemaps on a production project. Now that polymer-build is far enough along, it's time to add that support here.

When enabled, sourcemaps will be generated for any inline (or inlined) script tag that does not already have one. During the initial phase, the sourcemaps all contain generated mapping locations relative to the containing script tag. As a final set, these locations are updated so that they are absolute positions within the final bundle.

When debugging, http://sourcemapper.qfox.nl/ is your friend. It lets you easily see if the mappings are correct.

Once all the polymer tooling properly supports sourcemaps, then turning this flag to `true` by default is recommended. However until then they will just get in the way.

This PR depends on DefinitelyTyped/DefinitelyTyped/pull/15018 for type information for the `acorn.tokenizer` function.

I'm not super familiar with Typescript, so stylistically this may have issues. I'll be happy to correct them.


rictic:
Looks like the typings change has landed, kicked travis to rerun

src/import-utils.ts
```diff
@@ -217,6 +237,37 @@ export function rewriteAstBaseUrl(
...
+ * the script tag. Later this will be updated to account for the
+ * line offset within the final bundle.
+ */
+export async function addSourcemapsToInlineScripts(
+    document: Document, ast: ASTNode, oldBaseUrl: UrlString) {
+  const inlineScripts = dom5.queryAll(ast, matchers.inlineJavascript);
+  const promises = inlineScripts.map(scriptAst => {
+    let content = dom5.getTextContent(scriptAst);
+    const scriptLocation =
+        (scriptAst.__location as parse5.ElementLocationInfo).startTag;
```
rictic: There's a number of subtle issues with parse5's __locations that we handle in the analyzer. Recommend this instead:

```typescript
const parsedHtmlDocument = document.parsedDocument as ParsedHtmlDocument;
const sourceRange = parsedHtmlDocument.sourceRangeForStartTag(scriptAst)!;
```

That gets you a zero-indexed source range from the start to the end of the element's start tag. The biggest thing I notice that's relevant here is that it will handle newlines in the attribute section of a start tag correctly.

src/import-utils.ts
```diff
@@ -131,10 +144,17 @@ export async function inlineScript(document: Document, scriptTag: ASTNode) {
   }
 
   // Second argument 'true' tells encodeString to escape <script> tags.
-  const scriptContent =
-      encodeString(scriptImport.document.parsedDocument.contents, true);
+  let scriptContent = scriptImport.document.parsedDocument.contents;
+
+  if (enableSourcemaps) {
+    // It simplifies line calculation logic if inlined scripts don't
+    // start on the same line as the opening <script> tag
```
rictic: This comment reads like you're adding a newline to the beginning of the source but I don't see one here. Looks like you fixed the issue with the initial line in the source. Is this comment still relevant?

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as acorn from 'acorn';
```
rictic: Is it feasible to use espree instead? We already ship espree with the bundler, and it locks to a specific version of acorn, so I'm concerned that this will result in us shipping two copies of acorn.

Espree exposes its tokenizer I believe

src/test/souremap_test.ts
```diff
@@ -0,0 +1,82 @@
+/**
```
rictic: typo in file name. souremap_test -> sourcemap_test

ChadKillingsworth:
Bumped the acorn type definition version and merged in master to resolve the conflict. I can squash the commits if needed.

src/test/souremap_test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  suite('Sourcemaps', () => {
+
+    test('Basic', async() => {
+      const doc = await bundle(
+          'test/html/sourcemaps/A.html',
+          {inlineScripts: true, sourcemaps: true});
+      assert(doc);
+
+      const expected = [
```
rictic: Hm, is there a more semantic way of testing this? As is, this is a test will ensure that the output doesn't change, but it's not a legible or reviewable test of correctness.

A rough idea of the ideal here: 
 - get the range of the `console.log` statement before bundling
 - bundle
 - parse the sourcemap in the output with a standard library 
 - verify that the source map output maps the new location of console.log back to the old location

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+      newMapping.name = mapping.name;
+    }
+
+    generator.addMapping(newMapping);
+  });
+
+  return generator.toJSON();
+}
+
+export async function getExistingSourcemap(
```
rictic: This isn't used outside this file. For clarity, recommend against exporting it

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+const inlineSourcemapPrefix =
+    '\n//# sourceMappingURL=data:application/json;charset=utf8;base64,';
+
+const sourceMappingUrlExpr = /\n\/\/# sourceMappingURL=(.*)\n?/;
+
+/**
+ * Creates an identity source map from JS script content. Can offset original
+ * line/column data for inline script elements.
+ */
+export function createJsIdentitySourcemap(
```
rictic: This isn't used outside this file. For clarity, recommend against exporting it.

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  // Rewrite the sources array to be relative to the current URL
+  if (sourcemap.sources) {
+    sourcemap.sources =
+        sourcemap.sources.map(source => urlLib.resolve(mapUrl, source));
+  }
+  return sourcemap;
+}
+
+export async function addOrUpdateSourcemapComment(
```
rictic: Documentation please

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    return sourceContent + updatedSourcemapComment;
+  }
+}
+
+
+/**
+ * Update the source maps in inline script elements to reflect their relative
+ * line offset in the bundle.
```
rictic: Can you expand on this? e.g. from context it looks like this takes the AST of bundled output, finds inline source maps in script tags, and corrects them based on their new offsets in the HTML? Is that right?

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+  inlineScripts.forEach(script => {
+    let content = dom5.getTextContent(script);
+
+    const sourceMapContentParts = sourceMapCommentExpr.exec(content);
+    if (!sourceMapContentParts) {
+      return;
+    }
+    const scriptLocation =
+        (script.__location as parse5.ElementLocationInfo).startTag;
+    const sourceMapContent = Buffer.from(sourceMapContentParts[2], 'base64');
```
rictic: This logic is repeated a couple places, but it's somewhat noisy. Mind extracting it to a couple of module-local functions? i.e. two functions `(map: RawSourceMap) => string` and `(encodedMap: string) => RawSourceMap`.

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+    generator.addMapping(newMapping);
+  });
+
+  return generator.toJSON();
+}
+
+export async function getExistingSourcemap(
+    analyzer: AnalysisContext, sourceUrl: string, sourceContent: string) {
+  const dataUrlPrefix = 'data:application/json;';
+  const sourceMappingUrlParts = sourceMappingUrlExpr.exec(sourceContent);
```
rictic: Minor, and optional, but `exec` mutates the RegExp instance if it's global. These aren't global, so it's fine, but I needed to check to make sure this wouldn't have subtle regex state issues. I believe `.match` would serve just as well here, and it's more obviously safe.

src/source-map.ts
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Creates an identity source map from JS script content. Can offset original
+ * line/column data for inline script elements.
+ */
+export function createJsIdentitySourcemap(
+    sourceUrl: string,
+    sourceContent: string,
+    lineOffset: number = 0,
+    firstLineCharOffset: number = 0) {
```
rictic: We emit ES6 but we support Node v4 until it goes out of long term support in a month or so, so we can't use default parameters just yet. What we've been doing instead is:

```typescript
lineOffset?: number,
firstLineCharOffset?: number) {

if (lineOffset == null) {
  lineOffset = 0;
}
if (firstLineCharOffset == null) {
  firstLineCharOffset = 0;
}
```

It's kinda painful, but it's only for one more month!

src/import-utils.ts
ChadKillingsworth: What I need out of that though is a line/col number for the end of the `<script>` tag. Any way to obtain that from the analyzer?

src/source-map.ts
ChadKillingsworth: Probably - but now I get the joy of updating even more type definitions (because `tokenize` isn't in the espree definitions).

src/source-map.ts
ChadKillingsworth: I'm spoiled - Closure-compiler has long warned about any use of RegExp that depends on the global state. I don't even think about that any more ...

src/source-map.ts
ChadKillingsworth: Not quite. As this occurs during the import process, the final position of any token in the HTML document is unknown at this point. Instead, the mappings reflect the relative position within the script tag. It's not a correct source mapping yet - but an intermediate state.

src/test/souremap_test.ts
ChadKillingsworth: There about 100 better ways of doing this. I just was running short on time. I'll see what I can figure out.

src/source-map.ts
ChadKillingsworth: Typescript doesn't normalize for this? That's crazy.

src/import-utils.ts
rictic: The source range has three properties, filename, start, and end. Both start and end have line and column properties. So `sourceRange.end.line` and `sourceRange.end.column` will give you the end of the script tag.

src/source-map.ts
rictic: One thing we've found useful is defining a type locally in the custom_typings dir while developing, then contribute up to definitelytyped once you've got something working, but you don't have to block on it.

src/source-map.ts
rictic: Wow, that's a really nice lint rule to have. Especially if you can turn it off for the, like, 1% of cases where you actually do want to do exec() in a loop.

src/source-map.ts
rictic: You can emit ES5, or you can emit ES6. I think that's similar to what closure compiler supports? We're choosing ES6 and just limiting ourselves for now.

ChadKillingsworth:
I believe I addressed everything but the tests. Those will have to wait until at least tomorrow.

src/source-map.ts
ChadKillingsworth: I ended up using in the tests.

ChadKillingsworth:
Tests updated - and rebased onto master.

src/source-map.ts
ChadKillingsworth: I'm also pretty sure it will be needed for polymer-build.

ChadKillingsworth:
One question which has come up from me using this on real projects: what to do if an existing external source map can't be found or an inline source map contains invalid JSON or encoding?

As a dev, I want to know about this - but sometimes I can't correct it. For instance, in Polymer itself there's a reference to the ShadyCSS apply shim source map file.

It's easy to put try/catch blocks and create identity source maps in these cases. Just not sure the best way to notify the dev of an error (or if we should).

ChadKillingsworth:
@rictic I'm testing out these changes on my project and the espree change is throwing errors with ES6 template strings:

```
SyntaxError: Unexpected character '`'
```

rictic:
For espree you need to specify `ecmaVersion: 2017` in options and it will handle template strings, async await, etc.

src/source-map.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    return sourceContent + updatedSourcemapComment;
+  }
+}
+
+
+/**
+ * Update mappings in source maps within inline script elements to reflect
+ * their absolute position within a bundle. Assumes existing mappings
+ * are relative to their position within the script tag itself.
```
rictic: Is this assumption just due to where the source mapping takes place in the pipeline? i.e. that by the time it sees the source flowing through it's already been inlined so there's no way to distinguish external code that was inlined without updating its sourcemap from code that was e.g. in the original HTML and had its sourcemap?

Hm, and there's also the case of inlining some HTML which itself has inline JS which has a source map that has already been corrected for its place in that HTML file.

Asking just for clarity. Sourcemaps in inline javascript are rare enough, and this PR is already a significant enough chunk of work that we definitely shouldn't block on addressing it.

src/source-map.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    generator.addMapping(newMapping);
+  });
+
+  return generator.toJSON();
+}
+
+export async function getExistingSourcemap(
+    analyzer: AnalysisContext|Analyzer,
```
rictic: Can these methods just take an Analyzer? AnalysisContext is a private type in the analyzer and AFAIK there shouldn't be any way to get or use one directly without accessing private properties or directly importing its source file.

src/import-utils.ts
```diff
@@ -217,6 +236,35 @@ export function rewriteAstBaseUrl(
...
+ * the script tag. Later this will be updated to account for the
+ * line offset within the final bundle.
+ */
+export async function addSourcemapsToInlineScripts(
+    document: Document, ast: ASTNode, oldBaseUrl: UrlString) {
+  const inlineScripts = dom5.queryAll(ast, matchers.inlineJavascript);
+  const parsedHtmlDocument = document.parsedDocument as ParsedHtmlDocument;
+  const promises = inlineScripts.map(scriptAst => {
+    let content = dom5.getTextContent(scriptAst);
+    const sourceRange = parsedHtmlDocument.sourceRangeForStartTag(scriptAst)!;
```
rictic: Oh, hm, this is calling sourceRangeForStartTag for a Node that's from a reparsing of `parsedHtmlDocument.contents`. I'd recommend constructing a new `ParsedHtmlDocument` using properties off `document.parsedDocument` but with the fragment-parsed ast.

rictic:
The best way to handle invalid or not found source ranges I think would be to generate a Warning and propagate it up to polymer-build somehow. That way we could let the user know and draw a nice red squiggle underneath the line in question (though we'd probably want to truncate it to at most the first N bytes, to save the console users). 

I doubt such a channel exists at the moment though. I'd recommend a TODO for now, just to keep from delaying this PR or having it grow too large.

rictic:
BTW, sorry for the stream of nit picky comments, this code is super great, and I'm really excited to have real source map support in the bundler. We've wanted to produce source maps for a really long time!

src/source-map.ts
ChadKillingsworth: By design, this is the last step that occurs. All source maps for inline script tags at this point should have already been updated earlier to be relative to their script tag - regardless of how they originally existed.

> there's also the case of inlining some HTML which itself has inline JS which has a source map that has already been corrected for its place in that HTML file.

This shouldn't be possible. Inline script tags with sourcemaps are updated to relative positions as each import is processed. Tests have this case (I'll see about breaking out the tests to demonstrate each of these scenarios).

ChadKillingsworth:
Your comments are not nitpicky at all. I'd rather have a good review and get a solid PR merged. You are nowhere [even close to my record](https://github.com/google/closure-compiler/pull/2130).

src/source-map.ts
ChadKillingsworth: `AnalyzerContext` was coming from passing `document.analyzer` to this method. I suppose I could typecast that call ...

src/test/sourcemap_test.ts
```diff
@@ -0,0 +1,108 @@
...
+          {inlineScripts: true, sourcemaps: true, analyzer: analyzer});
+      assert(doc);
+      const compiledHtml = parse5.serialize(doc);
+
+      const inlineScripts = dom5.queryAll(doc, matchers.inlineJavascript);
+      assert.equal(inlineScripts.length, 6);
+
+      for (let i = 0; i < inlineScripts.length; i++) {
+        if (i === 5) {
+          continue;
```
ChadKillingsworth: I'll break up the tests - the last one doesn't have a console.log statement or a sourcemap (because it's not from an import).

ChadKillingsworth:
Switched back to espree - working on my project now.

I split apart / expanded the tests (and in the process found bugs). There are now tests that demonstrate all four scenarios:

 * Inline script without sourcemap
 * Inline script with sourcemap
 * External script without sourcemap
 * External script with sourcemap

If how this is working seems like quite the brain teaser - let me tell you it took me more than 2 weeks to figure out the original theory and do a proof of concept. This was compounded by the fact that I don't think I had ever seen a source map for an inline script and certainly couldn't find documentation for them. Discovering parse5 location info bugs didn't help matters ...

ChadKillingsworth:
@rictic Anything else needed for this?

src/import-utils.ts
rictic: ^

There's a decent chance that this is fine as is, but it's fragile to internal changes in the analyzer.

src/test/sourcemap_test.ts
```diff
@@ -0,0 +1,164 @@
...
+        const documents = await bundler.bundle([inputPath]);
+        return documents.get(inputPath)!.ast;
+      }
+
+  function getLine(original: string, lineNum: number) {
+    const lines = original.split('\n');
+    return lines[lineNum - 1];
+  }
+
+  async function testMapping(
```
rictic: Nice!

src/import-utils.ts
ChadKillingsworth: That one is being reparsed? I should just be reading the original AST there.

rictic: https://github.com/Polymer/polymer-bundler/pull/449/files#diff-66e48e66925dfeda3f9ec59c717ffd32R104 reparses it as a fragment

ChadKillingsworth: Hah - I was looking in my code for it. Totally missed it was already happening.

Fixed now!

src/import-utils.ts
```diff
@@ -107,8 +107,16 @@ export async function inlineHtmlImport(
   rewriteAstBaseUrl(importAst, resolvedImportUrl, document.url);
 
   if (enableSourcemaps) {
+    const reparsedDoc = new ParsedHtmlDocument({
+      url: document.url,
+      contents: htmlImport.document.parsedDocument.contents,
+      ast: importAst,
+      isInline: document.isInline,
```
rictic: Prefer `false` here over `document.isInline`, as this is definitely a top level doc, and without a location offset or an ast node it wouldn't be a valid inline doc anyways.

rictic:
I'm headed home for the night, will merge this in the morning!

rictic:
Awesome work! This will be going out in a new release ~today. We'll need to pipe it through the cli and polymer-build, but that work should be unblocked shortly.

PR #458, opened 2017-03-21
Excludes js

usergenic:
 - Excludes now honors JavaScript asset references:
   - Won't attempt to load the JS (which caused errors when local file not present.)
   - Won't inline excluded JS files.


usergenic:
I added a couple of tests, but this is really about bringing in @freshp86's PR into the 1.x branch.

PR #461, opened 2017-03-22
Add appveyor

usergenic:
- Added an appveyor.yml
 - `npm test` didn't work on windows because it invoked a posix-specific `npm run format` script.
 - I removed `npm run format` from the `npm test` script, because vscode can enforce clang.
 - Added `.vscode` in (copied from polymer-build) and removed `.vscode` from the `.gitignore`.
 - Changed a couple of path related things to path.posix to fix broken tests on windows.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,21 @@
+# Test against this version of Node.js
+environment:
+  nodejs_version: "4.8.0"
+
+# Install scripts. (runs after repo cloning)
+install:
+  # Get the latest stable version of Node.js or io.js
+  - ps: Install-Product node $env:nodejs_version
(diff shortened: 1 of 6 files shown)


appveyor.yml
```diff
@@ -0,0 +1,21 @@
...
+  # install modules
+  - npm install
+
+# Post-install test scripts.
+test_script:
+  # Output useful info for debugging.
+  - node --version
+  - npm --version
+  # run tests
+  - npm run test
```
rictic: `npm test` has better error messaging than `npm run test`

usergenic: That's good to know!  Fixin'

PR #464, opened 2017-03-30
Changed the Bundler.bundle method to take a manifest from Bundler.generateManifest

usergenic:
 - I updated README so it was at least more true than before.  It was woefully out-of-date.  Please don't review README for completeness as part of this branch, a broader README update is coming in subsequent PR.
 - This public API change is necessary to make it possible to identify the inlined content that should be filtered out of the build pipeline when bundling.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -67,6 +67,11 @@ export interface Options {
   stripExcludes?: UrlString[];
 }
 
+export interface BundleResult {
+  documents: DocumentCollection;
+  manifest: BundleManifest;
+}
+
 export class Bundler {
   analyzer: Analyzer;
   enableCssInlining: boolean;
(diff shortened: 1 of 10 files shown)


src/bundler.ts
```diff
@@ -130,6 +134,37 @@ export class Bundler {
   }
 
   /**
+   * Generates a BundleManifest with all bundles defined, using entrypoints,
+   * strategy and mapper.
+   *
+   * @param {Array<string>} entrypoints The list of entrypoints that will be
```
rictic: Recommend against including a type in the jsdoc. It's already in the typescript, and the typescript is checked, this is not. Can just leave it out and it's still legal jsdoc (I think).

usergenic: I'm all for leveraging the types from the TypeScript.  I've removed the type definitions.  Syntax highlighting for jsdoc starts to disappear so I'm using the hyphen to separate param name from description since it's valid jsdoc according to http://usejsdoc.org/tags-param.html and reads easier.

usergenic:
@rictic removed the type annotation from the jsdocs.  PTAL

usergenic:
Thanks :)

PR #463, opened 2017-03-30
fix _moveDomModuleStyleIntoTemplate

baocang:
- [x] fixed  https://github.com/Polymer/polymer-cli/issues/614

PR #466, opened 2017-04-07
Fix move style into template

usergenic:
 - The `_moveDomModuleStyleIntoTemplate` function did not get updated after `parse5` changed its handling of `<template>` content.  This updates bundler code to work with it correctly.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -370,10 +370,11 @@ export class Bundler {
     let template = dom5.query(domModule, matchers.template);
     if (!template) {
       template = dom5.constructors.element('template')!;
-      dom5.append(domModule, template);
+      template['content'] = dom5.constructors.fragment();
+      astUtils.prepend(domModule, template);
     }
     astUtils.removeElementAndNewline(style);
-    astUtils.prepend(template, style);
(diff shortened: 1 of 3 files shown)


usergenic:
@googlebot the History in https://github.com/Polymer/polymer-bundler/pull/463 demonstrates that @baocang intends their commits to be included in the project.

src/bundler.ts
```diff
@@ -370,10 +370,12 @@ export class Bundler {
     let template = dom5.query(domModule, matchers.template);
     if (!template) {
       template = dom5.constructors.element('template')!;
-      dom5.append(domModule, template);
+      treeAdapters.default.setTemplateContent(
+          template, dom5.constructors.fragment());
+      astUtils.prepend(domModule, template);
```
rictic: Interesting, what was this doing before?

usergenic: You can see the effects in one build condition here:
https://github.com/Polymer/polymer-cli/issues/614

rictic: So the template contents aren't serialized out at all?

usergenic: Yeah.

Issue #468, opened 2017-04-10
Change Bundler to work with Analyzer's InMemoryUrlLoader (required to upgrade Analyzer to 2.0.0-alpha.36)


usergenic:
Analyzer's `analyze()` method changed in the last release breaking Bundler's use of providing a serialized html document as second argument.

We need to put an `InMemoryUrlLoader` in the mix and feed it the file contents before calling `analyze()` on the filename to get Bundler's Analyzer to see the bundled version of the file.

PR #469, opened 2017-04-10
Use new InMemoryOverlayUrlLoader from `polymer-analyzer` to support `analyze` method breaking change

usergenic:
Analyzer dropped support for `analyze(url, contents)` but added an `InMemoryOverlayUrlLoader` as a convention to provide a map of urls to string contents.  This PR attempts to work with this new API while preserving the bundler's logic.

The approach is to create a private method on the `Bundler` itself to emulate the old `analyze` method signature and setup the overlay in the constructor.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -14,7 +14,8 @@
 import * as clone from 'clone';
 import * as dom5 from 'dom5';
 import {ASTNode, serialize, treeAdapters} from 'parse5';
-import {Analyzer, Document, FSUrlLoader} from 'polymer-analyzer';
+import * as path from 'path';
+import {Analyzer, Document, FSUrlLoader, InMemoryOverlayUrlLoader} from 'polymer-analyzer';
 
 import * as astUtils from './ast-utils';
 import * as bundleManifestLib from './bundle-manifest';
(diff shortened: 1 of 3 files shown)


usergenic:
Fixes https://github.com/Polymer/polymer-bundler/issues/468

src/bundler.ts
```diff
@@ -82,11 +83,29 @@ export class Bundler {
...
+    // In order for the bundler to use a given analyzer, we'll have to fork it
+    // so we can provide our own overlayUrlLoader which falls back to the
+    // analyzer's load method.
+    //
+    // TODO(usergenic): We can't delegate to `canLoad` directly, so we're
+    // proxying via the analyzer's `canResolveUrl` method.  We need to either
+    // expose `urlLoader` publicly on the analyzer or add a `canLoad` method.
+    if (opts.analyzer) {
+      this._overlayUrlLoader = new InMemoryOverlayUrlLoader({
+        canLoad: (url: string) => opts.analyzer!.canResolveUrl(url),
```
rictic: The not-null assertions are needed because `opts` is mutable, so just because analyzer exists here and now doesn't mean it will later. Recommend storing the analyzer in a local variable and then you can drop the `!`'s

```typescript
    if (opts.analyzer) {
      const analyzer = opts.analyzer;
      this._overlayUrlLoader = new InMemoryOverlayUrlLoader({
        canLoad: (url: string) => analyzer.canResolveUrl(url),
        load: async(url: string) => analyzer.load(url),
      });
```

src/bundler.ts
```diff
@@ -82,11 +83,29 @@ export class Bundler {
...
-        opts.analyzer :
-        new Analyzer({urlLoader: new FSUrlLoader()});
+
+    // In order for the bundler to use a given analyzer, we'll have to fork it
+    // so we can provide our own overlayUrlLoader which falls back to the
+    // analyzer's load method.
+    //
+    // TODO(usergenic): We can't delegate to `canLoad` directly, so we're
+    // proxying via the analyzer's `canResolveUrl` method.  We need to either
+    // expose `urlLoader` publicly on the analyzer or add a `canLoad` method.
```
rictic: Please file a bug for this, as canResolveUrl isn't a substitute for canLoad

rictic: Ah, this is done now, cool.

usergenic: Filed :)

src/bundler.ts
```diff
@@ -82,11 +83,30 @@ export class Bundler {
...
+    // https://github.com/Polymer/polymer-analyzer/issues/612
+    if (opts.analyzer) {
+      this._overlayUrlLoader = new InMemoryOverlayUrlLoader({
+        canLoad: (url: string) => opts.analyzer!.canResolveUrl(url),
```
rictic: Reposting comment that github decided was resolved (oh github):

The not-null assertions are needed because `opts` is mutable, so just because analyzer exists here and now doesn't mean it will later. Recommend storing the analyzer in a local variable and then you can drop the `!`'s

```typescript
    if (opts.analyzer) {
      const analyzer = opts.analyzer;
      this._overlayUrlLoader = new InMemoryOverlayUrlLoader({
        canLoad: (url: string) => analyzer.canResolveUrl(url),
        load: async(url: string) => analyzer.load(url),
      });
```

usergenic: Thank you for being my TypeScript mentor!  Done'd.

usergenic:
@rictic Taking it on like https://www.youtube.com/watch?v=djV11Xbc914

PR #472, opened 2017-04-10
Updated the vscode whitespace settings.

usergenic:
We need these settings everywhere.  Grabbed from `polymer-analyzer`.

### DIFF
--- a/src/test/bundler_test.ts
+++ b/src/test/bundler_test.ts
@@ -196,9 +196,9 @@ suite('Bundler', () => {
     const doc = await bundle('test/html/import-in-body.html');
     const imports = dom5.queryAll(doc, importMatcher);
     assert.equal(imports.length, 0);
-    const bodyContainer = dom5.query(doc, bodyContainerMatcher);
-    const scriptActual = dom5.query(doc, scriptExpected)!.parentNode;
-    const divActual = dom5.query(doc, divExpected)!.parentNode;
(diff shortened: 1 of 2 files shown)


usergenic:
@rictic suddenly type errors showing up in travis so I added some `!` to a test.

usergenic:
https://travis-ci.org/Polymer/polymer-bundler/builds/220755209 for reference.

src/test/bundler_test.ts
```diff
@@ -196,9 +196,9 @@ suite('Bundler', () => {
     const doc = await bundle('test/html/import-in-body.html');
     const imports = dom5.queryAll(doc, importMatcher);
     assert.equal(imports.length, 0);
-    const bodyContainer = dom5.query(doc, bodyContainerMatcher);
-    const scriptActual = dom5.query(doc, scriptExpected)!.parentNode;
-    const divActual = dom5.query(doc, divExpected)!.parentNode;
+    const bodyContainer = dom5.query(doc, bodyContainerMatcher)!;
```
rictic: This is the weirdest one. assert.equal requires non-null values?

usergenic: Yeah this just started.   Now anything cut from master is broken without the fix.

usergenic: Gracias!

PR #473, opened 2017-04-10
Use the new version of the Analyzer which implements UrlLoader.

usergenic:
Once Analyzer 2.0.0-alpha.37 is released, this code simplifies the constructor and fixes a bug where we incorrectly rely on `canResolveUrl` to proxy for `canLoad`.


usergenic:
Obviated by https://github.com/Polymer/polymer-bundler/pull/478

Issue #441, opened 2017-02-28
Make template URL fixup compatible with HTMLImports polyfill


sorvell:
The HTMLImports polyfill [has changed](https://github.com/webcomponents/html-imports/pull/38) how urls are re-written inside template elements and the bundler should do the same fix ups.

Now the polyfill only re-write url's inside templates that are (1) in the text of `<style>` elements or (2) in the `src` or `style` attributes of any element. 

This corresponds better to user expectations as these are inline loaded resources. It avoids, for example, re-writing url's in `href` attributes like `<a href>` which has been [an issue](https://github.com/Polymer/polymer/issues/4213) for Polymer users.

sorvell:
Let's hold off on this change for now. We need to figure out how Polymer's hybrid mode works with this before proceeding.

justinfagnani:
@sorvell, @kevinpschaaf, @usergenic and @valdrinkoshi what's the status of the URL rewriting changes? Do we have documentation on what rewriting is done, and what URL-related properties are available now?

kevinpschaaf:
There are docs in the migration guide [here](https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates).  Basically, now only URL's in CSS are rewritten to be relative to the import (e.g. `@import ...`, `background: url(...)`, etc.); **nothing else** (including e.g. `src`, `href` attributes) is ever rewritten now, and user should use `[[rootPath]]` or `[[importPath]]` bindings to determine what template paths are relative to.

@arthurevans Would it make sense to add a proper (non-migration-related) section re: paths in templates to [this page](https://www.polymer-project.org/2.0/docs/devguide/dom-template)?

The ability to do hybrid `[[rootPath]]` and `[[importPath]]` bindings was back-ported to 1.x in https://github.com/Polymer/polymer/pull/4373, but looks like we still need to make a `1.9` release to make that available. cc/ @azakus @sorvell to make sure we don't loose track of that.

usergenic:
To avoid surprising behaviors, this will be implemented as a global option to the bundler: `rewriteUrlAttrs` which will default to `false`.

usergenic:
@kevinpschaaf Just confirming-- does the polyfill still rewrites urls in elements' style attributes?

usergenic:
@sorvell I'm trying to get the bundler to mirror the expected behavior of the html imports polyfill wrt url rewriting in css, but I can't tell if this decision is settled.

[...]

PR #474, opened 2017-04-11
Don't rewrite template url attributes by default

usergenic:
Polymer 2.0 + HTML Imports polyfill etc doesn't rewrite urls inside `<template>` tags.  So Bundler won't unless a new `rewriteTemplateUrlAttrs` option is provided as `true`.

 - Fixes https://github.com/Polymer/polymer-bundler/issues/441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,13 +11,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   instead of entrypoints, strategy and mapper.  To produce a manifest,
   use new public method `generateManifest()`.
 - The polymer-bundler CLI now uses the current working directory as
-  the package root folder for its Analyzer, allowing absolute paths to 
+  the package root folder for its Analyzer, allowing absolute paths to
   resolve properly.
 - Fixed an issue where an immediate `<style>` child of `<dom-module>` was
   not moved into generated `<template>`.
-- BREAKING: Added option `rewriteTemplateUrls` to support old-style rewriting
-  of urls in element attributes for elements inside templates in inlined html
-  imports.  No longer does this by default.
+- BREAKING: Added option `rewriteUrlsInTemplates` to support rewriting of urls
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   use new public method `generateManifest()`.
 - Fixed an issue where an immediate `<style>` child of `<dom-module>` was
   not moved into generated `<template>`.
+- BREAKING: Added option `rewriteTemplateUrls` to support old-style rewriting
```
FredKSchott: could you explain what this means, for anyone who doesn't know about old-style?

usergenic: Good idea.  Done.  I also changed the name now to `rewriteUrlsInTemplates` which I think is more intention-revealing.

usergenic:
> Man, rewriteTemplateUrlAttrs really gets passed around from function to function. Would it be possible to support this functionality via a separate function call / mutation step instead?

In some ways it is a by-product of the dom5 API, being that query operations are provided an option to descend into `<template>` child nodes or not, which is essentially what `true` means in the case of `rewriteUrlsInTemplates` (formerly `rewriteTemplateUrlAttrs`).

While it is a flag that gets passed down, I think if we didn't work it that way, we'd need to basically just have conditionals and multiple additional call-sites to the various called rewrite functions in import-utils.

I agree that in the case of `inlineHtmlImport` its yet-another-param that feels like baggage and explosion of arity, but the behavior is significant and it can't be run after the inlining takes place.  In the case of functions like `rewriteAstToEmulateBaseTag` or `rewriteAstBaseUrl` I think it fits too.

usergenic:
@FredKSchott short answer: probably not without significant rework of the way these functions are composed and the benefit would be dubious.

usergenic:
per @valdrinkoshi comment https://github.com/Polymer/polymer-bundler/issues/441#issuecomment-293414941 I deem this PR mergeable as-is.

PR #477, opened 2017-04-12
Don't add html, head or body tags to bundled documents.

usergenic:
Don't add unwanted `html`, `head`, and `body` tags to bundled documents if they don't have them to begin with.

This fixes `polymer-bundler`'s contribution to the problem encountered here https://github.com/Polymer/polymer-build/issues/173 though `polymer-build`'s joiner contains cases where `polymer-build` still needs to be fixed.

### DIFF
--- a/src/ast-utils.ts
+++ b/src/ast-utils.ts
@@ -103,7 +103,6 @@ export function removeElementAndNewline(node: ASTNode, replacement?: ASTNode) {
   }
 }
 
-const injectedTagNames = new Set(['html', 'head', 'body']);
 /**
  * When parse5 parses an HTML document, it tries to fill in a few html tags
  * it considers missing if it doesn't see them (see `injectedTagNames` const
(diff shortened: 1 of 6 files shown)


src/ast-utils.ts
```diff
@@ -103,6 +103,39 @@ export function removeElementAndNewline(node: ASTNode, replacement?: ASTNode) {
...
+/**
+ * When parse5 parses an HTML document, it tries to fill in a few html tags
+ * it considers missing if it doesn't see them (see `injectedTagNames` const
+ * above.)  This function removes these elements from the AST so the AST
+ * represents only the html that was parsed.  The primary signal is that the
+ * node has no `__location` information, so this function can only reliably
+ * be used on a fresh parse, since subsequent tree manipulations may inject
+ * nodes without location information.
+ */
+export function removeFakeNodes(ast: dom5.Node) {
```
rictic: Please add a TODO to use the common version of this once it lands in either dom5 or the analyzer

PR #480, opened 2017-04-14
Spring cleaning

usergenic:
- Removed unimplemented options from the `polymer-bundler.js` CLI
 - Removed tests that didn't prove anything about those unimplemented options
 - Ran clang-format after pinning it to 1.0.49
 - Cleaned up minor README and CHANGELOG formatting issues.
 - Removed obsolete TODOs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,14 +113,14 @@ bundler.generateManifest([target]).then((manifest) => {
   bundler.bundle(manifest).then((bundles) => {
     /**
       * do stuff here
-      */      
+      */
   });
 });
 ```
 
 ## Caveats
 
-Because HTML Imports changes the order of execution scripts can have, polymer-bundler has to make a few compromises to achieve that same script 
+Because HTML Imports changes the order of execution scripts can have, polymer-bundler has to make a few compromises to achieve that same script
 execution order.
 
 1. Contents of all HTML Import documents will be moved to `<body>`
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -113,14 +113,14 @@ bundler.generateManifest([target]).then((manifest) => {
...
-      */      
+      */
   });
 });
 ```
 
 ## Caveats
 
-Because HTML Imports changes the order of execution scripts can have, polymer-bundler has to make a few compromises to achieve that same script 
+Because HTML Imports changes the order of execution scripts can have, polymer-bundler has to make a few compromises to achieve that same script
```
justinfagnani: I know this is a format change, but how do HTML imports change the order of execution of scripts?

src/test/sourcemap_test.ts
```diff
@@ -87,7 +87,7 @@ suite('Bundler', () => {
 
   suite('Sourcemaps', () => {
 
-    test('inline', async() => {
+    test('inline', async () => {
```
justinfagnani: while you're here, can the test names better describe what they're testing for, taking into account how mocha will print them, ie "Sourcemaps inline - passed"

README.md
usergenic: Fair point.  This was worded weirdly and wrongly-- This was meant to describe the problem of an html import forcing us to move imperatives like scripts and styles out of the head down into a hidden div after the inlined import. Rewritten.

src/test/sourcemap_test.ts
usergenic: Rewrite to try and be more intention revealing.

Issue #482, opened 2017-04-18
Need more test coverage for BundleUrlMapper scenarios


usergenic:
We are going to be exposing options for users to customize bundler strategies and url mappers.  If a BundleUrlMapper produces a url mapping for a bundle with one entrypoint that has a different path, there are a couple of issues to verify the behavior of with tests:

 - The bundle will not be *based* off the HTML structure of the initial entrypoint document, and may lose it's `<title>` attribute and/or other root element attributes.  We need this behavior spelled out in the tests or corrected in terms of loading from the entrypoint document as a basis.
 - The paths /may/ not be adjusted appropriately, where they were expected to be based off the entrypoint url.

Write tests for url mapper scenarios to demonstrate behavior and/or adjust sensibly.

Issue #437, opened 2017-02-24
Place shared_bundle_XX.html files at the right level and allow configuring it


web-padawan:
Today I tweaked `bundle.js` in my installed `polymer-build` module and started to play with `minEntrypoints` option to test how it works.

Here is the issue I faced: now, my shared bundles include paths like `src/components/app/my-app/my-app.html`, while real path to shell will be just `components/app/my-app/my-app.html`.

In short, it would be nice to have `shared_bundle_1.html` and others (actually I have 20 shared bundles for `minEntrypoints` = 3) placed into the separate directory rather than project root.

The reason is my app architecture: I believe that placing `bower_components` inside of the `src` is not a good idea and so I do some tricks. 

1. `index.html` is located inside of the `src` directory, while `bower_components` dir is outside:

```
root/
|
|--src/
|    |--components/
|    |--images/
|    |--index.html
`--bower_components/
```

2. For production, I have slightly different structure:
```
root/
|
|--components/
|--images/
|--bower_components/
`--index.html
```
 
3. To make this work, after bundling I use the following copy task to put them into `dist`:

```
function copyVulcanized () {
    return gulp.src([
       VULCANIZE_CACHE + '/src/components/app/**/*-app.html',
       VULCANIZE_CACHE + '/src/components/pages/**/*-page.html'
    ], { base: VULCANIZE_CACHE + '/src' })
    .pipe(gulp.dest(GULP_CACHE))
    .pipe($.size({
        title: 'vulcanize'
    }));
```
Those files listed via globs are shell and fragments (as of now, bundled build also contains lots of other stuff, see [this issue](https://github.com/Polymer/polymer-build/issues/70)).

In order to have shared bundles working, I need to rewrite paths somehow, because shell links, as well as `assetpath` attributes (and maybe other rewritten stuff too), are pointing to the `src` directory, which does not exist in production build.

But placing shared bundles in the right place would end up with relative paths like `../../my-app/my-app.html`, which is exactly what I want to reach.

What about making directory for shared bundles configurable? @usergenic WDYT?

/cc @FredKSchott as this path for shared bundles should be passed as some option from `polymer-project-config`, as well as `minEntrypoints` number.

justinfagnani:
[...]

PR #483, opened 2017-04-18
Bundle url mapper functions

usergenic:
https://github.com/Polymer/polymer-build/pull/189 opens the ability for users to provide options to the `PolymerProject#bundler()` method, opening up customized build strategies and opportunities.  One of the important elements of customization was allowing users to redefine where shared bundles go (https://github.com/Polymer/polymer-bundler/issues/437)

That behavior was hardwired into the default `BundleUrlMapper` used called `sharedBundleUrlMapper`. We were missing a friendly option for users to define their own bundle url mappers for shared bundles.  This adds those functions and factors out the layers of bundler url mapping.

Getting this released will also make it easier to write and read tests to demonstrate behavior of providing options added in https://github.com/Polymer/polymer-build/pull/189 as well as support more readable coverage per https://github.com/Polymer/polymer-bundler/issues/482.

 - [x] README.md has been updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ for this step).
 ## Options
 - `-h`|`--help`: print this message
 - `-v`|`--version`: print version number
-- `--exclude <path>`: exclude a subpath from root. Use multiple times to exclude multiple paths. Tags (imports/scripts/etc) that reference an excluded path are left in-place, meaning the resources are not inlined. ex: `--exclude=elements/x-foo.html --exclude=elements/x-bar.html`
+- `--exclude <path>`: exclude a URL. Use multiple times to exclude multiple paths. Tags (imports/scripts/etc) that reference an excluded path are left in-place, meaning the resources are not inlined. ex: `--exclude=elements/x-foo.html --exclude=elements/x-bar.html`
 - `--inline-scripts`: Inline external scripts.
 - `--inline-css`: Inline external stylesheets.
 - `--shell`: Uses a bundling strategy which puts inlines shared dependencies into a specified html app "shell".
@@ -87,6 +87,7 @@ polymer-bundler as a library has two exported function.
 
 `polymer-bundler` constructor takes an object of options similar to the command line options:
 
+- `analyzer`: An instance of `polymer-analyzer` that it will fork.  Bundler will create its own instance if this is not given.
 - `excludes`: URLs to exclude from inlining. URLs may represent files or folders. HTML tags referencing excluded URLs are preserved.
 - `inlineCss`: Inline external stylesheets.
 - `inlineScripts`: Inline external scripts.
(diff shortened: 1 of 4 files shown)


src/bundle-manifest.ts
```diff
@@ -286,19 +286,47 @@ function getBundleEntrypoint(bundle: Bundle): UrlString|null {
...
-    Map<UrlString, Bundle> {
-  let counter = 0;
-  const urlMap = new Map<UrlString, Bundle>();
-  for (const bundle of bundles) {
-    const bundleEntrypoint = getBundleEntrypoint(bundle);
-    if (bundleEntrypoint) {
-      urlMap.set(bundleEntrypoint, bundle);
-    } else {
-      urlMap.set(`shared_bundle_${++counter}.html`, bundle);
+export function generateSharedBundleUrlMapper(
```
FredKSchott: Create a type alias for the `SharedBundleUrlMapper` function signature to be the return value of this function

src/bundle-manifest.ts
```diff
@@ -286,19 +286,47 @@ function getBundleEntrypoint(bundle: Bundle): UrlString|null {
...
+        .forEach((url, i) => urlMap.set(url, sharedBundles[i]));
+    return urlMap;
+  };
+}
+
+/**
+ * Creates a bundle url mapper function which takes a prefix and appends an
+ * incrementing value, starting with `1` to the filename.
+ */
+export function generateSharedBundleCounterUrlMapper(urlPrefix: UrlString):
```
FredKSchott: If I understand correctly, this returns a common type of `SharedBundleUrlMapper`? This function name makes it sound like a different type of Mapper.

Consider renaming to `generateCountingSharedBundleUrlMapper()` or even `generatePrefixedSharedBundleUrlMapper()`.

README.md
```diff
@@ -97,23 +98,45 @@ polymer-bundler as a library has two exported function.
...
+var bundler = new require('polymer-bundler')();
+bundler.generateManifest(['my-app.html']).then((manifest) => {
+  bundle(manifest).then((bundles) => {
+    console.log('<!-- BUNDLED VERSION OF my-app.html: -->');
+    console.log(parse5.serialize(bundles.get('my-app.html').ast));
+  });
+});
+```
+
+An example, using options:
```
rictic: Consider:

`An example with a customized sharding strategy and output layout:`

src/bundle-manifest.ts
rictic: +1 to `generateCountingSharedBundleUrlMapper`

src/bundle-manifest.ts
usergenic: I am now explicitly declaring return type as BundleUrlMapper.

PR #484, opened 2017-04-18
Fixed bug where exclude did not support folders.

usergenic:
- Updated the code that filters out excluded html files to include the folder name case.
 - There was a test which was a false positive because it was actually a `<link rel="import" type="css" href="simple-style.css">` that was brought in from an inlined html file that was intended to be excluded.  Fixed the test.

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -282,6 +282,12 @@ export class Bundler {
     for (const bundle of bundles) {
       for (const exclude of this.excludes) {
         bundle.files.delete(exclude);
+        const excludeAsFolder = exclude.endsWith('/') ? exclude : exclude + '/';
+        for (const file of bundle.files) {
+          if (file.startsWith(excludeAsFolder)) {
+            bundle.files.delete(file);
+          }
+        }
       }
     }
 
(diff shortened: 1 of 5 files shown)


usergenic:
There was no issue for this.  I discovered the bug while a test failed after working on https://github.com/Polymer/polymer-bundler/pull/485 to inline scripts and css by default.

src/bundler.ts
```diff
@@ -282,6 +282,12 @@ export class Bundler {
     for (const bundle of bundles) {
       for (const exclude of this.excludes) {
         bundle.files.delete(exclude);
+        const excludeAsFolder = exclude.endsWith('/') ? exclude : exclude + '/';
```
aomarks: Doesn't seem like this would work under Windows -- should you use `path.sep` instead of `/`? Also, do we need to check for any special path operators? Maybe call `path.normalize` first?

usergenic: The exclude list matches urls in the files which have to be `/`

usergenic: I should verify its documentation makes such a claim.

usergenic: I mentioned URL only in one of the 3 places it was mentioned and the others said paths and files and subpaths -- So I've updated now, unifying the description as a URL.

aomarks: Ah ok, cool!

usergenic:
@aomarks thanks!

PR #485, opened 2017-04-18
Inline Scripts and CSS by default

usergenic:
In pretty much every example where bundler is used programmatically, we pass in `inlineCss` and `inlineScripts` as `true`.

It feels like we should just be doing this by default.

What do you think?

(Note: base of this PR is a branch with a fix that hasn't been merged to master yet for the `exclude-as-folder` behavior.  This is due to an tangential bug in the test that gets triggered when inlining is default behavior.)

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -90,8 +90,10 @@ export class Bundler {
 
     this.excludes = Array.isArray(opts.excludes) ? opts.excludes : [];
     this.stripComments = Boolean(opts.stripComments);
-    this.enableCssInlining = Boolean(opts.inlineCss);
-    this.enableScriptInlining = Boolean(opts.inlineScripts);
+    this.enableCssInlining =
+        Boolean(typeof opts.inlineCss === 'undefined' || opts.inlineCss);
+    this.enableScriptInlining = Boolean(
+        typeof opts.inlineScripts === 'undefined' || opts.inlineScripts);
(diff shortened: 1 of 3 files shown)


justinfagnani:
Yeah, this is what I would expect out of a bundler...

src/bundler.ts
```diff
@@ -90,8 +90,10 @@ export class Bundler {
 
     this.excludes = Array.isArray(opts.excludes) ? opts.excludes : [];
     this.stripComments = Boolean(opts.stripComments);
-    this.enableCssInlining = Boolean(opts.inlineCss);
-    this.enableScriptInlining = Boolean(opts.inlineScripts);
+    this.enableCssInlining =
+        Boolean(typeof opts.inlineCss === 'undefined' || opts.inlineCss);
```
justinfagnani: we can just do `opts.inlineCss === undefined`. We're not going to redefine undefined in any of our code. Setting `undefined` on the global does nothing, so we'd have to set it in one of our modules for this to be a problem.

justinfagnani: Also, this causes a boxing of the value. Can't we just do `options.inlineCss === undefined ? false : options.inlineCss` which matches what default parameters do in ES6

usergenic: You mean `options.inlineCss === undefined ? true : options.inlineCss` and also OK I will.

PR #486, opened 2017-04-18
Mutable manifest

usergenic:
- `BundleManifest` will now look up a `Bundle` by its `files` even after the `files` property has been mutated.  Supports post-processing of manifest which may be much simpler that embedding some kinds of manipulation inside `strategy` functions.  Notably, this enables users to emulate the old `add-import` behavior by updating a target bundle.
 - Sort of fixes https://github.com/Polymer/polymer-bundler/issues/402

### DIFF
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -58,7 +58,10 @@ export class AssignedBundle {
 }
 
 /**
- * A bundle manifest is a mapping of urls to bundles.
+ * A bundle manifest is a mapping of distinct urls to bundles.  A bundle's
+ * `files` property may be modified after the manifest is created, but it
+ * is always expected that no two bundles within the same manifest will
+ * contain the same file url.
  */
 export class BundleManifest {
   // Map of bundle url to bundle.
@@ -72,10 +75,13 @@ export class BundleManifest {
     this.bundles = urlMapper(Array.from(bundles));
   }
 
-  // Convenience method to return a bundle for a constituent file url.
+  /**
+   * Returns the url and bundle when a bundle is found containing the
+   * specified file url.
+   */
   getBundleForFile(fileUrl: UrlString): AssignedBundle|undefined {
     for (const [bundleUrl, bundle] of this.bundles) {
-      if (this.bundles.get(bundleUrl)!.files.has(fileUrl)) {
+      if (bundle.files.has(fileUrl)) {
         return {bundle: bundle, url: bundleUrl};
       }
     }
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -58,18 +58,21 @@ export class AssignedBundle {
 }
 
 /**
- * A bundle manifest is a mapping of distinct urls to bundles.  A bundle's
- * `files` property may be modified after the manifest is created, but it
- * is always expected that no two bundles within the same manifest will
- * contain the same file url.
+ * A bundle manifest is a mapping of distinct urls to bundles and acts as
+ * a specification for the bundle operation.
  */
 export class BundleManifest {
-  // Map of bundle url to bundle.
+  /**
+   * Map of bundle url to bundle. bundle's `files` property may be modified
+   * after the manifest is created, but it is always expected that no two
+   * bundles within the same manifest will contain the same file url.
+   */
   bundles: Map<UrlString, Bundle>;
 
   /**
-   * Given a collection of bundles and a BundleUrlMapper to generate urls for
-   * them, the constructor populates the `bundles` map.
+   * Given a collection of bundles and a `BundleUrlMapper` to generate urls
+   * for them, the constructor populates the `bundles` map.  Files in a
+   * bundle should be unique to that bundle.
    */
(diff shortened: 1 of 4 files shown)


src/bundle-manifest.ts
```diff
@@ -64,32 +64,20 @@ export class BundleManifest {
   // Map of bundle url to bundle.
   bundles: Map<UrlString, Bundle>;
```
rictic: If external writes to this map are allowed then it seems like we should document the contract a bit more.

When is it acceptable to mutate these bundles? May a file be in multiple bundles?

src/bundle-manifest.ts
```diff
@@ -64,32 +64,20 @@ export class BundleManifest {
...
   }
 
   // Convenience method to return a bundle for a constituent file url.
-  getBundleForFile(url: UrlString): AssignedBundle|undefined {
-    const bundleUrl = this._bundleUrlForFile.get(url);
-    if (bundleUrl) {
-      return {bundle: this.bundles.get(bundleUrl)!, url: bundleUrl};
+  getBundleForFile(fileUrl: UrlString): AssignedBundle|undefined {
+    for (const [bundleUrl, bundle] of this.bundles) {
+      if (this.bundles.get(bundleUrl)!.files.has(fileUrl)) {
```
rictic: `this.bundles.get(bundleUrl)!` is just `bundle` yeah?

usergenic: yes it returns a `Bundle`

usergenic: oh lol

usergenic: I AM SMUART

usergenic: fixed.

src/bundle-manifest.ts
usergenic: Updated comments on manifest.

justinfagnani:
Do we really want to make manifest mutable, or do we want to make it easy to make cloned and modified manifests at the right place in the bundle pipeline?

usergenic:
@justinfagnani Since the bundler doesn't mutate the manifest while it's working with it, I suppose we could preserve the expectation of immutability within Bundler by cloning inside the call to `bundle()`.

My reason for why the object should be mutable is because we'd have to otherwise add a whole host of methods to do simple changes like adding/moving/removing files and/or bundles to the manifest and have them return clones or whatever-- it just gets a little unwieldy and expensive to develop that API.  The benefit to the manifest is to having a plan for what the bundler should do and making it easy for people to adjust that plan in their scripts is where we want to be.

usergenic:
I clone the manifest inside `bundle()` now, preserving the given values.

justinfagnani:
@usergenic where would the user be modifying the manifest? In a custom strategy?

usergenic:
Initial motivation was to provide a simple option for the now missing vulcanize option `add-import`.

Right now, all manipulations of bundles occurs in the bundle strategy and crafting a custom bundle strategy to add/remove specific files to specific bundles is doable.  In this way mutable manifests aren't *necessary*.

The thinking with making manifest mutable was that ad hoc specific adjustments to the bundles after they've been mapped to their urls etc in the manifest would be potentially a more ergonomic option for certain circumstances and users.  For example, I know I want to add a file or remove a file in production build and I can do that in a very intentional way with the manifest, whereas it is a little fiddlier if done in a custom strategy even with the composeStrategies combinator. 

Side note, there's not *yet* a good affordance for manipulating the manifest inside a polymer-build powered gulpfile.  A script that used polymer-bundler directly could do this after receiving the manifest from the call to `generateManifest()` and before sending it off to `bundle()`.

One *possible* means to incorporate this to `PolymerProject#bundler()` might be a callback option in the soon-to-be-released options like `project.bundler({prepareManifest: (manifest) => {...}})`.

justinfagnani:
Thanks, I'm just trying to see where and how this would be used.

So it seems like this potential callback: `project.bundler({prepareManifest: (manifest) => {...}})` is very nearly like a strategy, except that it would be expected to mutate its argument.

Wouldn't that be essentially solved by https://github.com/Polymer/polymer-build/pull/189 where you can already do: `project.bundler({strategy: (bundles: Bundle[]) => {...}})`. A strategy that returns a new `Bundle[]` seems relatively easy to write. You could copy the array and mutate that, you could call `.map`, `.filter`., etc.

One problem is that a strategy doesn't have access to the bundle URLs that are in the manifest, so a user might have to write a BudleStrategy and a BundleUrlMapper. Maybe those aren't useful as separate interfaces and we should combine them, have that take a manifest, and allow a user to specify a _chain_ of strategies so that they can easily inject a transformation/renaming on top of built-in strategies:

`project.bundler({transforms: [sharedDepsStrategy(3), shellStrategy('shell.html'), (manifest: BundleManifest) => { ...; return new BundleManifest(...); }})`

justinfagnani:
Also, if this makes a manifest just a wrapper on a `Map<UrlString, Bundle>` can we just get rid of the manifest class and pass around a the map and turn the methods and constructor into functions?

Issue #447, opened 2017-03-06
support for lazy-imports


cordovapolymer:
Currently polymer-bundler doesn't support https://github.com/Polymer/lazy-imports/ which leads to inability to build such project with polymer-cli.
Is there any hope that such `<link rel="lazy-import" href="lazy-element.html" group="lazy">` imports will be bundled?

I've found a pull request https://github.com/Polymer/polymer-bundler/pull/391 for this, but it looks like incomplete.

web-padawan:
I think you should list all lazy impors as fragments in `polymer.json` and they will be bundled.

cordovapolymer:
@web-padawan , that's obvious but that negates the benefits of lazy-imports against importHref.

justinfagnani:
@usergenic let's talk about this on Monday and see if you need any additional support from the Analyzer. One thing could be an option to easily say whether a URL is only ever used as a lazy import, which would make it an entrypoint, or maybe an API to get a list of all inferred entrypoints.

jay8t6:
+1

stramel:
@usergenic I have been trying to create a minimal reproduction repo but am failing pretty hard. Let me reach out to my co-workers that are having the issue and see if they can work on it. Will let you know what comes of it.

usergenic:
I wrote tests last night to identify the current behavior with lazy imports and they're effectively treated the same as eager imports, regardless of what the code thinks it's trying to do...

I've also started on the fix to make purely lazy edges (i.e an import which is only ever imported lazily) convert to entrypoints (bundlers term for an edge or fragment).  I should hopefully have a bundler PR ready for it today.

jay8t6:
I have added a repo to replicate this issue.

stramel:
Re: @jay8t6's comment, there was some confusion around which issue he was having, this seems like it is more analyzer/build. I will link it to an issue over there.

usergenic:
Just released in `polymer-bundler` 2.0.0-pre.13.

(Should get into release pipeline for `polymer-build` and `polymer-cli` soon.)

stramel:
Thanks @usergenic!

abdonrd:
Yay! Thanks @usergenic! Excited with the lazy imports support! 😁

stramel:
Seems something is still broken here... https://github.com/stramel/lazy-import-bundle

It looked like it was working fine when I was lazily importing pages.

usergenic:
[...]

PR #487, opened 2017-04-25
Lazy imports up in here

usergenic:
Rewrote `buildDepsIndex` to do-the-right-thing™ wrt `lazy-import`.

- BREAKING: Bundler now supports "lazy imports" aka `<link rel="lazy-import">`.
  Imports which are lazy-imported are now implicitly treated as entrypoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- BREAKING: Bundler now supports "lazy imports" aka `<link rel="lazy-import">`.
+  URLs which are imported lazily are implicitly treated as entrypoints.  This
+  is a breaking change because, previously, Bundler would incorrectly treat
+  lazy imports just like regular imports.
 - BREAKING: Bundler now inlines Scripts and CSS by default.  Pass `inlineCss`
   and `inlineScripts` as `false` explicitly to disable.
 - Fixed an issue where `exclude` option did not actually support folder names.
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -113,8 +113,7 @@ export class Bundler {
     for (const bundleEntry of manifest.bundles) {
       const bundleUrl = bundleEntry[0];
       const bundle = {url: bundleUrl, bundle: bundleEntry[1]};
-      const bundledAst =
-          await this._bundleDocument(bundle, manifest, bundle.bundle.files);
+      const bundledAst = await this._bundleDocument(bundle, manifest);
       bundledDocuments.set(
           bundleUrl, {ast: bundledAst, files: Array.from(bundle.bundle.files)});
     }
@@ -227,10 +226,8 @@ export class Bundler {
    */
   private async _bundleDocument(
       docBundle: AssignedBundle,
-      bundleManifest: BundleManifest,
-      bundleImports?: Set<string>): Promise<ASTNode> {
+      bundleManifest: BundleManifest): Promise<ASTNode> {
     let document = await this._prepareBundleDocument(docBundle);
-
     const ast = clone(document.parsedDocument.ast);
     dom5.removeFakeRootElements(ast);
     this._appendHtmlImportsForBundle(ast, docBundle);
--- a/src/deps-index.ts
+++ b/src/deps-index.ts
@@ -11,7 +11,6 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
-import {AssertionError} from 'assert';
 import {Analyzer, Document} from 'polymer-analyzer';
 import {UrlString} from './url-utils';
 
@@ -21,79 +20,89 @@ export interface DepsIndex {
 }
 
 type DependencyMapEntry = {
-  url: UrlString,
-  // Eagerly reachable dependencies.
-  eager: Set<UrlString>,
-  // Lazily reachable dependencies - may also include some entries from "eager"
-  lazy: Set<UrlString>
+  // All dependencies of the document
+  deps: Set<UrlString>,
+  // Eagerly loaded dependencies of the document
+  eagerDeps: Set<UrlString>,
+  // All imports defined with `<link rel="lazy-import">`
+  lazyImports: Set<UrlString>,
 };
 
-async function _getTransitiveDependencies(
-    url: UrlString, entrypoints: UrlString[], analyzer: Analyzer):
-    Promise<DependencyMapEntry> {
-      const analysis = await analyzer.analyze([url]);
-      const document = analysis.getDocument(url);
-      if (!(document instanceof Document)) {
-        const message = document && document.message || 'unknown';
-        throw new Error(`Unable to get document ${url}: ${message}`);
-      }
-      const imports = document.getFeatures(
-          {kind: 'import', externalPackages: true, imported: true});
-      const eagerImports = new Set<UrlString>();
-      const lazyImports = new Set<UrlString>();
-      for (const htmlImport of imports) {
-        try {
-          console.assert(
-              htmlImport.url, 'htmlImport: %s has no url', htmlImport);
-        } catch (err) {
-          if (err instanceof AssertionError) {
-            continue;
-          }
-          throw err;
-        }
+/**
+ * For a given document, return a set of transitive dependencies, including
+ * all eagerly-loaded dependencies and lazy html imports encountered.
+ */
+function getHtmlDependencies(document: Document): DependencyMapEntry {
+  const deps = new Set<UrlString>();
+  const eagerDeps = new Set<UrlString>();
+  const lazyImports = new Set<UrlString>();
+  _getHtmlDependencies(document, true, deps, eagerDeps, lazyImports);
+  return {deps, eagerDeps, lazyImports};
+}
 
-        switch (htmlImport.type) {
-          case 'html-import':
-            eagerImports.add(htmlImport.document.url);
-            break;
-          case 'lazy-html-import':
-            lazyImports.add(htmlImport.document.url);
-            break;
-        }
-      }
-      return {url: url, eager: eagerImports, lazy: lazyImports};
+function _getHtmlDependencies(
+    document: Document,
+    viaEager: boolean,
+    visited: Set<UrlString>,
+    visitedEager: Set<UrlString>,
+    lazyImports: Set<UrlString>) {
+  const htmlImports = document.getFeatures(
+      {kind: 'html-import', imported: false, externalPackages: true});
+  for (const htmlImport of htmlImports) {
+    const importUrl = htmlImport.document.url;
+    if (htmlImport.lazy) {
+      lazyImports.add(importUrl);
     }
-
-export async function buildDepsIndex(
-    entrypoints: UrlString[], analyzer: Analyzer): Promise<DepsIndex> {
-  const entrypointToDependencies: Map<UrlString, Set<UrlString>> = new Map();
-  const dependenciesToEntrypoints: Map<UrlString, Set<UrlString>> = new Map();
-  const queue = Array.from(entrypoints);
-  const visitedEntrypoints = new Set<UrlString>();
-  while (queue.length > 0) {
-    const entrypoint = queue.shift()!;
-    if (visitedEntrypoints.has(entrypoint)) {
+    if (visitedEager.has(importUrl)) {
       continue;
     }
-    const dependencyEntry =
-        await _getTransitiveDependencies(entrypoint, entrypoints, analyzer);
-    const dependencies = new Set(dependencyEntry.eager);
-    dependencies.add(entrypoint);
-    entrypointToDependencies.set(entrypoint, dependencies);
-    for (const lazyDependency of dependencyEntry.lazy.values()) {
-      if (!visitedEntrypoints.has(lazyDependency)) {
-        queue.push(lazyDependency);
+    const isEager = viaEager && !htmlImport.lazy;
+    if (isEager) {
+      visitedEager.add(importUrl);
+      // In this case we've visited a node eagerly for the first time,
+      // so recurse
+    } else if (visited.has(importUrl)) {
+      // In this case we're seeing a node lazily again, so don't recurse
+      continue;
+    }
+    visited.add(importUrl);
+    _getHtmlDependencies(
+        htmlImport.document, isEager, visited, visitedEager, lazyImports);
+  }
+}
+
+/**
+ * Analyzes all entrypoints and determines each of their transitive
+ * dependencies.
+ * @param entrypoints Urls of entrypoints to analyze.
+ * @param analyzer
+ * @return a dependency index of every entrypoint, including entrypoints that
+ *     were discovered as lazy entrypoints in the graph.
+ */
+export async function buildDepsIndex(
+    entrypoints: UrlString[], analyzer: Analyzer): Promise<DepsIndex> {
+  const depsIndex = {entrypointToDeps: new Map<UrlString, Set<UrlString>>()};
+  const analysis = await analyzer.analyze(entrypoints);
+  const allEntrypoints = new Set<UrlString>(entrypoints);
+
+  // Note: the following iteration takes place over a Set which may be added
+  // to from within the loop.
+  for (const entrypoint of allEntrypoints) {
+    const documentOrWarning = analysis.getDocument(entrypoint);
+    if (documentOrWarning instanceof Document) {
+      const deps = getHtmlDependencies(documentOrWarning);
+      depsIndex.entrypointToDeps.set(
+          entrypoint, new Set([entrypoint, ...deps.eagerDeps]));
+      // Add lazy imports to the set of all entrypoints, which supports
+      // recursive
+      for (const dep of deps.lazyImports) {
+        allEntrypoints.add(dep);
       }
+    } else {
+      const message = documentOrWarning && documentOrWarning.message || '';
+      console.warn(`Unable to get document ${entrypoint}: ${message}`);
     }
   }
-  entrypointToDependencies.forEach((dependencies, entrypoint, map) => {
-    for (const dependency of dependencies) {
-      if (!dependenciesToEntrypoints.has(dependency)) {
-        dependenciesToEntrypoints.set(dependency, new Set());
-      }
-      const entrypointSet = dependenciesToEntrypoints.get(dependency)!;
-      entrypointSet.add(entrypoint);
-    }
-  });
-  return {entrypointToDeps: entrypointToDependencies};
+
+  return depsIndex;
 }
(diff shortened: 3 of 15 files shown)


usergenic:
Fixes https://github.com/Polymer/polymer-bundler/issues/447

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
-            break;
-          case 'lazy-html-import':
-            lazyImports.add(htmlImport.document.url);
-            break;
-        }
-      }
-      return {url: url, eager: eagerImports, lazy: lazyImports};
+  // Gets a collection of only the direct html imports for the document.
+  const htmlImports = document.getFeatures(
+      {kind: 'html-import', imported: false, externalPackages: true});
```
rictic: hm, I think this should be named `getTransitiveHtmlDependencies` then

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+ * For a given URL, return a set of eager and lazy dependencies.  The
+ * function is recursive, but a `visited` set is carried forward to short-
+ * circuit loops encountered.
+ */
+function getTransitiveDependencies(
+    url: UrlString,
+    analysis: Analysis,
+    visited: Set<UrlString> = new Set<UrlString>()): DependencyMapEntry {
+  const map: DependencyMapEntry = {
+    eager: new Set(),
```
FredKSchott: helpful to add a type here. Set() of what?

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
-      const entrypointSet = dependenciesToEntrypoints.get(dependency)!;
-      entrypointSet.add(entrypoint);
-    }
+  const depsIndex = {entrypointToDeps: new Map<UrlString, Set<UrlString>>()};
+  const analysis = await analyzer.analyze(entrypoints);
+  const processed = new Set<UrlString>();
+  const unprocessed = new Set<UrlString>(entrypoints);
+  unprocessed.forEach((entrypoint) => {
+    processed.add(entrypoint);
+    unprocessed.delete(entrypoint);
```
FredKSchott: oof, editting the object being iterated over always gives me the creeps. Iterate over `entrypoints` or `new Set(entrypoints)` here instead

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+    const deps = getTransitiveDependencies(entrypoint, analysis);
+    depsIndex.entrypointToDeps.set(
+        entrypoint, new Set([entrypoint].concat(Array.from(deps.eager))));
+
+    // Add lazy edges as unprocessed entrypoints, if they haven't already been
+    // processed.
+    deps.lazyEdges.forEach((dep) => processed.has(dep) || unprocessed.add(dep));
```
FredKSchott: can you expand this to be more clear? new lines are free :)

src/import-utils.ts
```diff
@@ -93,8 +105,13 @@ export async function inlineHtmlImport(
     return;
   }
 
-  // If the analyzer could not load the import document, we can't inline it, so
-  // lets skip it.
+  // We will stop short of actually inlining a `lazy-import` or its children.
```
FredKSchott: Why? Would be useful to know for future devs

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+
+  // Every unvisited direct lazy dependency's dependencies are recognized
+  // as transitively lazy unless already defined as eager.
+  Array.from(map.lazy).filter((d) => !visited.has(d)).forEach((d) => {
+    const deps = getTransitiveDependencies(d, analysis, visited);
+    deps.eager.forEach(
+        (dep) => map.eager.has(dep) ? map.lazy.delete(dep) : map.lazy.add(dep));
+    deps.lazy.forEach((dep) => map.eager.has(dep) || map.lazy.add(dep));
+    deps.lazyEdges.forEach((dep) => map.lazyEdges.add(dep));
+  });
```
rictic: This algorithm isn't obviously correct to me. Here's my thinking based on what I've inferred from what we're trying to do here, lemmy know if I go off track:

A file has eager and lazy imports of other files. A file is an eager dependency of another if that file is reachable by any path of entirely eager imports. A file is a lazy dependency of another if all paths to that file include at least one lazy import.

This means that every eager dependency of an eager dependency is an eager dependency, and the rest of its dependencies are lazy. So I think an easier to understand implementation would involve building the set of files reachable by by entirely eager paths, and another set of all files reachable at all, then constructing the lazy files as those that are in the set of all files but not in the set of eager files.

I think this algorithm is equivalent in complexity but simpler to reason about.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- BREAKING: Bundler now supports "lazy imports" aka `<link rel="lazy-import">`.
```
justinfagnani: Breaking or new feature?

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+    url: UrlString,
+    analysis: Analysis,
+    visited: Set<UrlString> = new Set<UrlString>()): DependencyMapEntry {
+  const map: DependencyMapEntry = {
+    eager: new Set(),
+    lazy: new Set(),
+    lazyEdges: new Set()
+  };
+  const document = analysis.getDocument(url);
+  if (!(document instanceof Document)) {
```
justinfagnani: we do this all over right now, because getDocument() may return a warning

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
-        dependenciesToEntrypoints.set(dependency, new Set());
-      }
-      const entrypointSet = dependenciesToEntrypoints.get(dependency)!;
-      entrypointSet.add(entrypoint);
-    }
+  const depsIndex = {entrypointToDeps: new Map<UrlString, Set<UrlString>>()};
+  const analysis = await analyzer.analyze(entrypoints);
+  const processed = new Set<UrlString>();
+  const unprocessed = new Set<UrlString>(entrypoints);
+  unprocessed.forEach((entrypoint) => {
```
rictic: prefer `for ... of` over `.forEach`

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+    processed.add(entrypoint);
+    unprocessed.delete(entrypoint);
+    const deps = getTransitiveDependencies(entrypoint, analysis);
+    depsIndex.entrypointToDeps.set(
+        entrypoint, new Set([entrypoint].concat(Array.from(deps.eager))));
```
rictic: golf: `[entrypoint].concat(Array.from(deps.eager))` -> `[entrypoint, ...deps.eager]`

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+ * function is recursive, but a `visited` set is carried forward to short-
+ * circuit loops encountered.
+ */
+function getTransitiveDependencies(
+    url: UrlString,
+    analysis: Analysis,
+    visited: Set<UrlString> = new Set<UrlString>()): DependencyMapEntry {
+  const map: DependencyMapEntry = {
+    eager: new Set(),
+    lazy: new Set(),
```
rictic: is `lazy` ever used? or is that for a future PR?

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
-      if (!dependenciesToEntrypoints.has(dependency)) {
-        dependenciesToEntrypoints.set(dependency, new Set());
-      }
-      const entrypointSet = dependenciesToEntrypoints.get(dependency)!;
-      entrypointSet.add(entrypoint);
-    }
+  const depsIndex = {entrypointToDeps: new Map<UrlString, Set<UrlString>>()};
+  const analysis = await analyzer.analyze(entrypoints);
+  const processed = new Set<UrlString>();
+  const unprocessed = new Set<UrlString>(entrypoints);
```
rictic: Because it's a set, if you only added entries to `unprocessed` and never deleted then you could get rid of `processed` entirely. Could rename it from `unprocessed` to just `urls`

src/deps-index.ts
usergenic: Well, I'm iterating over the set I'm mutating because I want to add things to it like a queue as in https://github.com/Polymer/polymer-bundler/pull/487#discussion-diff-113536810R116

src/deps-index.ts
justinfagnani: Some light afternoon reading: https://www.kent.edu/sites/default/files/TR-KSU-CS-2010-01.pdf

We should be able to do this in a single BFS, with a change that a node can be visited more than once if we're seeing it via an eager edge for the first time, essentially performing two reachability queries at the same time:

Something like this modified BFS should work:

```js
getHtmlDependencies(document) {
  const deps = new Set();
  const eagerDeps = new Set();
  const lazyImports = new Set();
  _getHtmlDependencies(document, true, deps, eagerDeps, lazyImports);
  return {deps, eagerDeps, lazyImports};
}

_getHtmlDependencies(document, viaEager, visited, visitedEager, lazyImports) {
  const imports = document.getFeatures({kind: 'html-import', imported: false, externalPackages: true});
  for (const import of imports) {
    const importUrl = import.document.url;
    if (import.lazy) {
      lazyImports.add(importUrl);
    }
    if (visitedEager.has(importUrl)) {
      continue;
    }
    const isEager = viaEager || !import.lazy;
    if (isEager) {
      visitedEager.add(importUrl);
      // In this case we've visited a node eagerly for the first time,
      // so recurse
    } else if (visited.has(importUrl)) {
      // In this case we're seeing a node lazily again, so don't recurse
      continue;
    }
    visited.add(importUrl);
    getHtmlDependencies(import.document, isEager, visited, visitedEager, lazyImports);
  }
}
```

src/deps-index.ts
usergenic: I *kinda* want an `analysis.getDocumentOrBust(url);` option to dry up these redundant go-like error-handling clauses.

src/deps-index.ts
usergenic: unprocessed is really a queue.  it's added to within the loop that iterates over it.  i'll look at rewriting to make that more obvious.

usergenic: (its a queue that doesn't care about order.)

rictic: I got that, but you don't need `processed`, and you don't need to delete from `unprocessed`

src/deps-index.ts
usergenic: i'm replacing all of them.

src/deps-index.ts
usergenic: +1

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+  htmlImports.forEach((i) => {
+    if (i.lazy) {
+      map.lazy.add(i.document.url);
+      map.lazyEdges.add(i.document.url);
+    } else {
+      map.eager.add(i.document.url);
     }
+  });
+
+  visited.add(url);
```
usergenic: moot, because replaced with your awsm code

src/deps-index.ts
```diff
@@ -21,79 +21,100 @@ export interface DepsIndex {
...
+
+  // Every unvisited direct eager dependency's lazy dependencies are recognized
+  // as transitively lazy unless already recognized as eager.  All others are
+  // treated as eager.
+  Array.from(map.eager).filter((d) => !visited.has(d)).forEach((d) => {
+    const deps = getTransitiveDependencies(d, analysis, visited);
+    deps.eager.forEach((dep) => map.eager.add(dep) && map.lazy.delete(dep));
+    deps.lazy.forEach((dep) => map.eager.has(dep) || map.lazy.add(dep));
```
usergenic: this has been deleted in favor of new shorter algorithm.

src/deps-index.ts
usergenic: I totally used this.

src/deps-index.ts
usergenic: Done.  Renamed to `allEntrypoints`

src/deps-index.ts
usergenic: replaced all of them.

src/deps-index.ts
usergenic: LIKE.  Using.

src/deps-index.ts
usergenic: Now that I don't use processed and unprocessed, this is simpler:
```typescript
    // Add lazy imports as entrypoints to index.
    for (const dep of deps.lazyImports) {
      allEntrypoints.add(dep);
    }
  ```

File: src/import-utils.ts (hunk 5 above)
Comment (usergenic) [inline src/import-utils.ts]: Updated comment:

```typescript
  // We don't actually inline a `lazy-import` because its loading is intended
  // to be deferred until the client requests it.
```

src/deps-index.ts
usergenic: this is gone.  replaced with new type.

src/deps-index.ts
usergenic: Done.

CHANGELOG.md
usergenic: Updated CHANGELOG comment to explain why it was BREAKING.

src/bundler.ts
```diff
@@ -227,10 +226,8 @@ export class Bundler {
    */
   private async _bundleDocument(
       docBundle: AssignedBundle,
-      bundleManifest: BundleManifest,
-      bundleImports?: Set<string>): Promise<ASTNode> {
```
usergenic: the tslint file has `        "no-unused-variable": true,`

src/deps-index.ts
```diff
@@ -21,79 +20,86 @@ export interface DepsIndex {
...
-  const visitedEntrypoints = new Set<UrlString>();
-  while (queue.length > 0) {
-    const entrypoint = queue.shift()!;
-    if (visitedEntrypoints.has(entrypoint)) {
-      continue;
+  const depsIndex = {entrypointToDeps: new Map<UrlString, Set<UrlString>>()};
+  const analysis = await analyzer.analyze(entrypoints);
+  const allEntrypoints = new Set<UrlString>(entrypoints);
+  for (const entrypoint of allEntrypoints) {
+    const document = analysis.getDocument(entrypoint);
```
justinfagnani: maybe name `documentOrWarning`

src/deps-index.ts
```diff
@@ -21,79 +20,86 @@ export interface DepsIndex {
...
+    if (!(document instanceof Document)) {
+      const message = document && document.message || '';
+      throw new Error(`Unable to get document ${entrypoint}: ${message}`);
```
justinfagnani: I don't think we should throw in this case. Ideally we only throw for program errors, and bad user input is handled in normal execution and warnings and errors are threaded through the results. Here we should continue building if possible and just warn and exit with a non-0 code. It would be a bit of work to do that now, probably changing the return type and such, so maybe we can just `console.warn` for now.

usergenic:
@justinfagnani @rictic @FredKSchott Addressed all PR feedback.  Also, gutted and replaced the `getTransitiveDependencies` method with the more efficient one from Justin.

src/deps-index.ts
```diff
@@ -21,79 +20,86 @@ export interface DepsIndex {
...
-        queue.push(lazyDependency);
-      }
+    const deps = getHtmlDependencies(document);
+
+    depsIndex.entrypointToDeps.set(
+        entrypoint, new Set([entrypoint, ...deps.eagerDeps]));
+
+    // Add lazy imports as entrypoints to index.
+    for (const dep of deps.lazyImports) {
+      allEntrypoints.add(dep);
```
justinfagnani: does this cause the for loop to iterate over the lazy import? Do we need to process it? I'd consider making this explicit by making a queue of entrypoints and adding the lay import to the queue.

src/import-utils.ts
```diff
@@ -77,11 +79,22 @@ export async function inlineHtmlImport(
...
-    // means we've handled that entire bundle, so we guard against inlining any
-    // other file from that bundle by checking the visited urls for the bundle
-    // url itself.
-    if (visitedUrls.has(importBundle.url)) {
+    if (
+        // If we've previously visited a url that is part of another bundle, it
+        // means we've handled that entire bundle, so we guard against inlining
+        // any other file from that bundle by checking the visited urls for the
+        // bundle url itself.
+        visitedUrls.has(importBundle.url) &&
```
justinfagnani: consider making a well-named variable to hold this boolean value to summarize what it indicates. I like the comments :)

src/test/deps-index_test.ts
```diff
@@ -23,66 +23,61 @@ import {buildDepsIndex} from '../deps-index';
...
+      const lazyImport2 = 'lazy-imports/lazy-import-2.html';
+      const shared1 = 'lazy-imports/shared-eager-import-1.html';
+      const shared2 = 'lazy-imports/shared-eager-import-2.html';
+      const shared3 = 'lazy-imports/shared-eager-and-lazy-import-1.html';
+      const analyzer =
+          new Analyzer({urlLoader: new FSUrlLoader('test/html/imports')});
+      const expectedEntrypointsToDeps = new Map([
+        [entrypoint, new Set([entrypoint])],
+        [lazyImport1, new Set([lazyImport1, shared1, shared2])],
+        [lazyImport2, new Set([lazyImport2, shared1, shared2, shared3])],
```
justinfagnani: I'd like to see a test with a lazy import that itself has a lazy import that has dependencies, so test whether or not lazy imports need to be fully processed as entrypoints.

src/bundler.ts
rictic: tsconfig is the place to do this

usergenic: Ok.  Adding in separate PR so it doesn't muddy this one. (there are a few other spots including exported functions with unused parameters.)

usergenic: https://github.com/Polymer/polymer-bundler/issues/490

src/deps-index.ts
usergenic: Ok.  Just console.warn'n for now.

src/deps-index.ts
usergenic: Done.

src/deps-index.ts
usergenic: I added a comment to call it out, since it was surprising (and awesome).

src/import-utils.ts
usergenic: Ok.  Did it!

src/test/deps-index_test.ts
usergenic: I know we said in-person that we didn't need this but I added it anyways.  Test upgraded with lazily imported imports of lazy imports.

usergenic:
Okay @justinfagnani lazy imports PR feedback has been addressed.  PTAL!

Issue #493, opened 2017-05-02
Export urlMappers and strategy generators on default namespace


usergenic:
Make the strategy and url mappers available in default namespace so users don't have to things out of `polymer-bundler/lib/bundle-manifest`.

usergenic:
Released in polymer-bundler 2.0.0-pre.14

PR #495, opened 2017-05-03
Exposing bundle url mapper and strategy generator functions right on the polymer-bundler module.

usergenic:
- Unexported a couple of methods that are really just utility functions for the existing strategies etc.
 - Sorted the various functions to make it easier to understand what's available.
 - Got rid of the `sharedBundleUrlMapper` constant.  Better to use the generate commands and not have confusing variant sitting around.


usergenic:
Fixes https://github.com/Polymer/polymer-bundler/issues/493

usergenic:
cc @FredKSchott in-particular.

usergenic:
:-D Thanks, @rictic

PR #496, opened 2017-05-03
Moved strategy and urlMapper out of the generateManifest function arguments and into the constructor and properties of Bundler.

usergenic:
- This is an ergonomic thing that I think ultimately makes it easier to understand how to use Bundler configuration options inside `polymer-build` too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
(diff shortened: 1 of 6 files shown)


usergenic:
This PR has some opportunistic updates to the README.  I don't want to get too hung up on that because I need a larger README overhaul PR, but I need to document that the strategy and urlMapper options moved.  I incorporated some of the suggestions from the update-readme PR from earlier.  Will get to more of them soon.  I think its important to get these changes in though so that `polymer-build` release can include them and its README can make more sense in context (since these options are now all going to be in one place instead of split across constructor and generateManifest.)

usergenic:
Also please note this is based off my `export-stuff` branch and won't go in until that's merged in.

Issue #494, opened 2017-05-03
lazy-import link tags are being moved around like regular link tags


usergenic:
Symptom: your element has some lazy imports inside its dom-module.  After running bundler, they're moved out of the dom-module (BAD!)

This is probably because the code that moves imperatives into the hidden div on import is not differentiating lazy imports from other imports.

usergenic:
@stramel thanks for reporting this

stramel:
**Before:** 
![image](https://cloud.githubusercontent.com/assets/855184/25646266/61746a44-2f7c-11e7-824e-4d6c57fe28a7.png)

**After build:** 
![image](https://cloud.githubusercontent.com/assets/855184/25646283/8832c28e-2f7c-11e7-8980-402e662416e4.png)

---

**Steps to Reproduce:**

1. Clone `https://github.com/stramel/lazy-import-bundle` (master branch, should be default)
2. `npm i;bower i`
3. `npm run build`
4. `polymer serve build/ -o`
5. Navigate to "View Two" from the side nav

**Expected Result:** Vaadin Grid displayed with list of demo data (This can be viewed by serving the non-build version)

**Actual Result:** Vaadin Grid is not displayed or loaded

usergenic:
Fixed in #497 now in master.

stramel:
@usergenic  Still not fixed with latest changes.

![image](https://cloud.githubusercontent.com/assets/855184/25684099/285926d8-3025-11e7-9517-b78616f89154.png)

usergenic:
Reopening until we merge the latest fix #499

usergenic:
#499 is merged now.  Closing again; will go in next release.

stramel:
:+1:

PR #497, opened 2017-05-03
Fixed issue where lazy-import tags were being moved around just like eager import tags.

usergenic:
- The code that moves around links into the hidden div was including lazy-import links because they are html imports.  That was bad and has been fixed.


usergenic:
cc @stramel

This fixes #494

PR #499, opened 2017-05-03
Fixed case where an inlined import's lazy-import tags were being moved.

usergenic:
- Fixed case where an inlined import's `<link rel="lazy-import">` tags were being moved.

Yeah.

### DIFF
--- a/src/test/bundler_test.ts
+++ b/src/test/bundler_test.ts
@@ -177,14 +177,36 @@ suite('Bundler', () => {
   });
 
   test('lazy imports are not moved', async () => {
-    assert.equal(
-        dom5.queryAll(
-                await bundle('test/html/imports/lazy-imports.html'),
-                preds.AND(
-                    preds.parentMatches(preds.hasTagName('dom-module')),
-                    matchers.htmlImport))
-            .length,
-        2);
+    const bundler = new Bundler({
+      analyzer:
+          new Analyzer({urlLoader: new FSUrlLoader('test/html/imports')})
+    });
+    const manifest = await bundler.generateManifest(['lazy-imports.html']);
+    const documents = await bundler.bundle(manifest);
+    const matchLazyImportInsideDomModule = preds.AND(
+        preds.parentMatches(preds.hasTagName('dom-module')),
+        matchers.htmlImport);
+
+    // The `lazy-imports.html` file has 3 imports in the head of the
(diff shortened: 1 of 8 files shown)


usergenic:
@stramel

Fixes https://github.com/Polymer/polymer-bundler/issues/494 some more.  (Okay previous fix only fixed one case and that was not actually the case reported by the bug)

src/test/bundler_test.ts
```diff
@@ -177,14 +177,36 @@ suite('Bundler', () => {
...
-        2);
+    const bundler = new Bundler({
+      analyzer:
+          new Analyzer({urlLoader: new FSUrlLoader('test/html/imports')})
+    });
+    const manifest = await bundler.generateManifest(['lazy-imports.html']);
+    const documents = await bundler.bundle(manifest);
+    const matchLazyImportInsideDomModule = preds.AND(
+        preds.parentMatches(preds.hasTagName('dom-module')),
+        matchers.htmlImport);
```
justinfagnani: should this match all imports or just lazy?

src/test/bundler_test.ts
```diff
@@ -177,14 +177,36 @@ suite('Bundler', () => {
...
+
+    // The `lazy-imports.html` file has 3 imports in the head of the
+    // document.  The first is eager and should be moved.  The remaining
+    // two are lazy imports and should not be moved.
+    const entrypointBundle = documents.get('lazy-imports.html')!.ast;
+    console.log(parse5.serialize(entrypointBundle));
+    const entrypointLazyImports = dom5.queryAll(
+        entrypointBundle,
+        preds.AND(preds.parentMatches(matchers.head), matchers.htmlImport));
```
justinfagnani: same here

usergenic: In the `<head>` case here, I do want to match all html imports so I can verify the eager one is removed by verifying the number of imports matching is 2 instead of 3.

src/test/bundler_test.ts
usergenic: I think *I* was lazy here.  The matcher here isn't even used until after the first query, so this is defined prematurely, and also, the fact that it matches all html imports isn't clear from the name, which really should just be "matchHtmlImportsInDomModule".  I'm getting rid of the variable altogether and being specific in the actual query now.

PR #500, opened 2017-05-04
Updated README samples to actually be valid.

usergenic:
The sample code in the README was _quite_ incorrect.  I tried running it.  Now I've updated it so it at least runs.

💯


usergenic:
Thanks, @aomarks

PR #508, opened 2017-05-11
Updated the `bundle()` method to return a `BundleResult`.

usergenic:
This PR is an important prerequisite that will make it possible for `polymer-build` to filter out inlined files from its `BuildBundler` transform stream.

 - This result contains a clone of the provided manifest, where
   each Bundle contains a record of the imports it inlined.

### DIFF
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -12,6 +12,8 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import * as clone from 'clone';
+
 import {UrlString} from './url-utils';
 
 /**
@@ -42,6 +44,11 @@ export class Bundle {
 
   // Set of all files included in the bundle.
   files: Set<UrlString>;
+
+  // These sets are updated as bundling occurs.
+  inlinedHtmlImports = new Set<UrlString>();
+  inlinedScripts = new Set<UrlString>();
+  inlinedStyles = new Set<UrlString>();
 
   constructor(entrypoints?: Set<UrlString>, files?: Set<UrlString>) {
     this.entrypoints = entrypoints || new Set<UrlString>();
@@ -83,6 +90,11 @@ export class BundleManifest {
         this._bundleUrlForFile.set(fileUrl, bundleUrl);
       }
     }
+  }
+
+  // Returns a clone of the manifest.
+  clone(): BundleManifest {
+    return clone(this);
   }
 
   // Convenience method to return a bundle for a constituent file url.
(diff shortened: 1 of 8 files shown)


src/bundle-manifest.ts
```diff
@@ -248,13 +260,22 @@ export function generateShellMergeStrategy(
...
+       } of bundles) {
+    newBundle.entrypoints =
+        new Set<UrlString>([...newBundle.entrypoints, ...entrypoints]);
+    newBundle.files = new Set<UrlString>([...newBundle.files, ...files]);
+    newBundle.inlinedHtmlImports = new Set<UrlString>(
+        [...newBundle.inlinedHtmlImports, ...inlinedHtmlImports]);
+    newBundle.inlinedScripts =
+        new Set<UrlString>([...newBundle.inlinedScripts, ...inlinedScripts]);
+    newBundle.inlinedStyles =
+        new Set<UrlString>([...newBundle.inlinedStyles, ...inlinedStyles]);
```
usergenic: I wish `Set#add()` was multi-argument like `Array#push()`

src/bundle-manifest.ts
```diff
@@ -85,6 +92,11 @@ export class BundleManifest {
     }
   }
 
+  // Returns a clone of the manifest.
+  clone(): BundleManifest {
```
rictic: Maybe `fork()` would be more consistent with our naming elsewhere?

src/test/bundler_test.ts
```diff
@@ -44,12 +49,23 @@ suite('Bundler', () => {
...
       }
 
   suite('Default Options', () => {
 
+    test('URLs for inlined HTML imports are recorded in Bundle', async () => {
+      await bundle(inputPath);
+      assert.sameMembers(
+          [...documentBundle.inlinedHtmlImports],
+          ['imports/simple-import.html'],
+          [...documentBundle.inlinedHtmlImports].join(', '));
```
rictic: Here and elsewhere, why this third option? Shouldn't `assert.sameMembers` give a good error message for the missing elements without this? We may need to ask chai to display a diff, I think there's a method you can call

usergenic: sameMembers gave me a terrible error message when I ran it:

[Array(2)] doesn't have the same members as [Array(2)]

usergenic: I'll try it with sameDeepMembers

usergenic: sameMembers:
`AssertionError: expected [ Array(3) ] to have the same members as [ Array(2) ]`

sameDeepMembers:
`AssertionError: wtf: expected [ Array(3) ] to have the same members as [ Array(2) ]`

usergenic: I could remove it and make the developer to add a console.log when the test starts failing.  This was just an attempt to save that step.

WDYT?

src/bundle-manifest.ts
usergenic: Ok.

src/test/bundler_test.ts
rictic: Definitely no on the console.log. Looking around in the chai repo, someone [may have added diffing to sameMembers](https://github.com/chaijs/chai/pull/863)? Two more things to try, and if they don't work your solution is good:

1) Update chai to latest version
2) Set `chai.config.showDiff = true;`

usergenic: I converted to deepEqual with a sort for the multi-member set.  Just as short as sameMembers but removes the message arguments.

Issue #507, opened 2017-05-11
[question] Is hidden div required for bundles not containing html?


web-padawan:
In our project we have special imports used to de-duplicate libraries like `d3` and `ace` and to put them to shared bundles. After using crisper, our bundle looks like this:
```html
<html>

<head></head>

<body>
    <div hidden="" by-polymer-bundler=""></div>
    <script src="bundle-7.js"></script>
</body>

</html>
```
That empty `div` used to contain one `script` tag before running `crisper`.
Can we just remove that div in those cases? Should bundler catch this or not?

usergenic:
I'm all for bundler removing the hidden div if it's empty.  Great idea.

web-padawan:
Actually this div only becomes empty after extracting script tag. Maybe it would be enough to check that bundle contains nothing but inlined scripts.

PR #514, opened 2017-05-19
Do not include the `by-polymer-bundler` hidden div in bundled files if it is empty.

usergenic:
- This is seemingly innocuous, but it helps simplify integration tests up in the CLI.
 - Also fixes https://github.com/Polymer/polymer-bundler/issues/507

### DIFF
--- a/src/bundler.ts
+++ b/src/bundler.ts
@@ -489,7 +489,7 @@ export class Bundler {
   }
 
   /**
-   * Removed all empty hidden container divs from the AST.
+   * Removes all empty hidden container divs from the AST.
    */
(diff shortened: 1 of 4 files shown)


src/bundler.ts
```diff
@@ -484,4 +487,15 @@ export class Bundler {
     dom5.removeFakeRootElements(ast);
     return this._analyzeContents(document.url, serialize(ast));
   }
+
+  /**
+   * Removed all empty hidden container divs from the AST.
```
aomarks: s/Removed/Remove

PR #518, opened 2017-05-23
Add the sourcesContent field to generated source maps

Lange:
Hello! :wave:

When working with the source maps generated by the `polymer-bundler`, I realized that although they gave me accurate stack traces, clicking on those traces would bring up a blank source in DevTools. After a bit of research, I leanred that this was because the source maps generated by `polymer-bundler` lacked a `sourcesContent` field, which is what DevTools uses to actually render the original source.

I am new to the science behind source maps (and to TypeScript!), so please forgive me if I've made some naive mistakes here.

I also have not yet added tests for this change in functionality, because I couldn't quite figure out how. Again, pardon my lack of experience.

If this is a desired change, then I'd be happy to put some more work into writing tests that cover this change.

Thanks! 🎉 

### DIFF
--- a/src/source-map.ts
+++ b/src/source-map.ts
@@ -72,6 +72,7 @@ function createJsIdentitySourcemap(
     generator.addMapping(mapping);
   });
 
+  generator.setSourceContent(sourceUrl, sourceContent);
   return generator.toJSON();
 }
 
@@ -96,6 +97,10 @@ function offsetSourceMap(
     }
 
     generator.addMapping(newMapping);
+  });
+
+  sourcemap.sources.forEach(source => {
+    generator.setSourceContent(source, consumer.sourceContentFor(source));
   });
 
   return generator.toJSON();
--- a/src/source-map.ts
+++ b/src/source-map.ts
@@ -46,6 +46,7 @@ function rawSourceMapToBase64String(sourcemap: RawSourceMap) {
 export function createJsIdentitySourcemap(
     sourceUrl: string,
     sourceContent: string,
+    originalFileContent: string,
     lineOffset: number,
     firstLineCharOffset: number) {
   const generator = new SourceMapGenerator();
@@ -72,7 +73,7 @@ export function createJsIdentitySourcemap(
     generator.addMapping(mapping);
   });
 
-  generator.setSourceContent(sourceUrl, sourceContent);
+  generator.setSourceContent(sourceUrl, originalFileContent);
   return generator.toJSON();
 }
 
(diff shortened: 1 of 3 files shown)


src/source-map.ts
```diff
@@ -72,6 +72,7 @@ function createJsIdentitySourcemap(
     generator.addMapping(mapping);
   });
 
+  generator.setSourceContent(sourceUrl, sourceContent);
```
Lange: Looks like there's cases where there is a newline char (`\n`) at the beginning of the `sourceContent` here that causes the resulting map to be off by one line. I don't know of a simple check for `if (sourceContent.startsWith('\n'))` would be adequate? Might generate false positives.

Lange:
Upon further investigation, the root of this issue might be that the path to certain relative sources is being resolved incorrectly. I think that when the path is resolved correctly, the file is automatically loaded from disk into `sourcesContent`.

Lange:
Ah nope, this is indeed just a straight up bug. The `sourcesContent` is dropped from the sourcemap unless it is explicitly carried over each time, like this PR does.

Lange:
I have added two tests for this. I'm unsure if the route I took is the way that you'd want these to be implemented, but they do seem to do the job (they fail on `master`, pass on this PR).

The problem described in [this comment](https://github.com/Polymer/polymer-bundler/pull/518#discussion_r118114108) is still present though, and should be resolved before this PR is merged. I think I need input from a maintainer on how to go about doing that.

usergenic:
@Lange looking at this now

src/source-map.ts
usergenic: Not sure if related yet, but I found this interesting: https://www.html5rocks.com/en/tutorials/developertools/sourcemaps/#toc-xssi

```js
if (response.slice(0, 3) === ")]}") {
    response = response.substring(response.indexOf('\n'));
}
```

Lange:
@usergenic I'm having some weird issues unrelated to either of my open pull requests. Is it okay if I DM you on the Polymer slack?

Lange:
Oh boy, this definitely needs more testing before being merged. I just tried it out when my script was inlined instead of external and the resulting map is pretty wild and off by several lines:
image

Lange:
I removed my previous tests and have replaced them with a check that the `sourcesContent` in every test is an exact match for the actual file on disk. This causes pretty much every test to fail, and reveals that there is more work to be done.

Lange:
Got it!

image

This should be good to go, pending review.

Lange:
_sigh_ Naturally, I have spoken too soon. I've discovered one more case where it will still generate an invalid mapping. Working on it.

usergenic:
@Lange I'm going to merge this into the include-sources-content branch and continue working on this.  Let me know if you have other insights, not-yet-applied.  And thanks again for the patch and great details on this PR.

Lange:
Hi @usergenic, that sounds good! I don't have any further insights to share at this time -- I've had to focus on other tasks. Thanks!

PR #519, opened 2017-05-24
Don't add the `original` property to a mapping unless one exists

### LINKED ISSUE #516: Sourcemaps don't work
I was trying to write my own build build script instead of using `polymer bundle`. So I wrote something that looks like this ...
``` js
const parse5 = require('parse5');
const {Bundler} = require('polymer-bundler');

const bundler = new Bundler({
  sourcemaps: true,
});

bundler.generateManifest(['index.html']).then((manifest) => {
  bundler.bundle(manifest).then((result) => {
    const parsedHtml = parse5.serialize(result.documents.get('index.html').ast);
    // ...
  }).catch(console.error);
}).catch(console.error);
```
... as specified in the docs.

But it outputs an error something like this
```
Error: Invalid mapping: {"generated":{"line":2,"column":0},"source":null,"original":{"line":null,"column":null},"name":null}
    at SourceMapGenerator_validateMapping [as _validateMapping] (/mnt/d/web/testing/sandboxes/polymer/node_modules/source-map/lib/source-map-generator.js:277:13)
    at SourceMapGenerator_addMapping [as addMapping] (/mnt/d/web/testing/sandboxes/polymer/node_modules/source-map/lib/source-map-generator.js:101:12)
    at consumer.eachMapping.mapping (/mnt/d/web/testing/sandboxes/polymer/node_modules/polymer-bundler/lib/source-map.js:84:19)
    at Array.forEach (native)
    at BasicSourceMapConsumer.SourceMapConsumer_eachMapping [as eachMapping] (/mnt/d/web/testing/sandboxes/polymer/node_modules/source-map/lib/source-map-consumer.js:155:14)
    at offsetSourceMap (/mnt/d/web/testing/sandboxes/polymer/node_modules/polymer-bundler/lib/source-map.js:71:14)
    at Object.<anonymous> (/mnt/d/web/testing/sandboxes/polymer/node_modules/polymer-bundler/lib/source-map.js:141:25)
    at Generator.next (<anonymous>)
    at fulfilled (/mnt/d/web/testing/sandboxes/polymer/node_modules/polymer-bundler/lib/source-map.js:17:58)
```

If the `sourcemaps` options is either set to `false` or removed completely, everything works fine. And there is not enough documentation on how to extract generated sourcemap from the `result` object.

Is this a bug, or is there something I am missing, please let me know.

Lange:
This fixes an error that can be encountered when generating source maps. This is needed for mappings to pass the `_validateMapping` check in [source-map-generator.js#L264](https://github.com/mozilla/source-map/blob/master/lib/source-map-generator.js#L264)

This passes the existing tests, but I'm unsure how to add a new test to ensure that this won't happen again in the future.

Fixes #516

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+- Fixed an error when generating source maps
 <!-- Add new, unreleased changes here. -->
 
 ## 2.0.1 - 2017-05-19
--- a/src/source-map.ts
+++ b/src/source-map.ts
@@ -83,13 +83,16 @@ function offsetSourceMap(
   consumer.eachMapping(mapping => {
     const newMapping: any = {
       source: mapping.source,
-      original: {line: mapping.originalLine, column: mapping.originalColumn},
       generated: {
         line: mapping.generatedLine + lineOffset,
         column: mapping.generatedColumn +
             (mapping.generatedLine === 1 ? firstLineCharOffset : 0)
       }
     };
+
+    if (typeof mapping.originalLine === 'number' && typeof mapping.originalColumn === 'number') {
+      newMapping.original = { line: mapping.originalLine, column: mapping.originalColumn };
+    }
 
     if (mapping.name) {
       newMapping.name = mapping.name;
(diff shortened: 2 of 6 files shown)


Lange:
I have added a test for this which fails on `master` but passes on this PR.

Interestingly, this same test passes if the HTML and the JS files are in the same directory. It seems to be only when the two files are in different directories that this failure occurs. This may indicate another, separate issue.

usergenic:
Will review this today, @Lange - Thanks for your submission!

PR #524, opened 2017-05-28
Remove es6-promise from package.json.

### LINKED ISSUE #523: package.json pulls es6-promise which is not used by polymer-bundler.
es6-promise is not used by the code but is listed in package.json.  This causes polymer-build to install two copies of es6-promise - the old version for this module and the current version for sw-precache.

coreyfarrell:
Fixes #523


justinfagnani:
Thanks!

Issue #517, opened 2017-05-22
Inlines are in the wrong order if a fragment is also eagerly imported.


justinfagnani:
This is from a customer issue we debugged at I/O.

The scenario is app.html imports home.html with imports utils.html. home.html is specified as a fragment though, and then gets inlined before utils.html, causing runtime errors. Removing home.html as a fragment fixes the issue.

usergenic:
Repro:
```html
<!-- shared-thing.html -->
<script>'I should be inlined before fragment-a';</script>
```
```html
<!-- fragment-a.html -->
<link rel='import' href='shared-thing.html'>
```
```html
<!-- shell.html -->
<link rel='import' href='fragment-a.html'>
```
Bundled:
```html
<!-- shell.html -->
<link rel='import' href='fragment-a.html'>
<script>'I should be inlined before fragment-a';</script>
```

This occurs because shared things are injected into the shell.  Currently, they are blithely appended to the shell body, but this is problematic in the case where a non-inlined eager-import is part of the shell and depends on the thing being injected (in this case `fragment-a.html` depends on `shared-thing.html` coming before it.

A solution is to prepend inlined unvisited dependencies of the eagerly imported bundles before their links.

PR #528, opened 2017-06-02
Fix issue with ordering of html imports injected by shell strategy.

usergenic:
- [x] CHANGELOG updated.
- Fixes https://github.com/Polymer/polymer-bundler/issues/517
- Added bundler tests to demonstrate where the problem exists when injecting html imports doesn't consider the order they are needed.
- See the test fail here https://github.com/Polymer/polymer-bundler/pull/528/commits/8f23344f421cd50d5f12fc60f8964c726ef46368
- See the test pass here https://github.com/Polymer/polymer-bundler/pull/528/commits/d234caedc3fc74bcd68a440c2118a27688971b74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+- When injecting html imports into documents for inlining, bundler now tries
+  to put them ahead of imports which will depend on them instead of just
+  appending to end of document. This is especially important in shell strategy
+  apps with other shared bundles or when shell eagerly imports fragments.
 <!-- Add new, unreleased changes here. -->
 
 ## 2.0.1 - 2017-05-19
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-- When injecting html imports into documents for inlining, bundler now tries
-  to put them ahead of imports which will depend on them instead of just
-  appending to end of document. This is especially important in shell strategy
-  apps with other shared bundles or when shell eagerly imports fragments.
+- In cases such as the shell merge strategy, imports which are not originally
+  part of the shell are now inserted before other imports which are dependent
+  on them.  Fixes https://github.com/Polymer/polymer-bundler/issues/517.
 <!-- Add new, unreleased changes here. -->
 
 ## 2.0.1 - 2017-05-19
--- a/src/ast-utils.ts
+++ b/src/ast-utils.ts
@@ -114,6 +114,28 @@ export function parse(html: string, options: ParserOptions): ASTNode {
 }
 
 /**
+ * Returns true if the nodes are given in order as they appear in the source
+ * code.
+ */
+export function inOrder(left: ASTNode, right: ASTNode): boolean {
+  const l = left.__location, r = right.__location;
+  return l && r && l['line'] && r['line'] &&
+      (l['line'] < r['line'] ||
+       (l['line'] === r['line'] && l['col'] < r['col']));
+}
+
+/**
+ * Returns true if both nodes have the same line and column according to their
+ * location info.
+ */
+export function sameNode(node1: ASTNode, node2: ASTNode): boolean {
+  const l1 = node1.__location, l2 = node2.__location;
+  return !!(
+      l1 && l2 && l1['line'] && l1['col'] && l1['line'] === l2['line'] &&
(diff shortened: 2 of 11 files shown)


src/test/deps-index_test.ts
```diff
@@ -89,5 +90,26 @@ suite('Bundler', () => {
           serializeMap(index.entrypointToDeps),
           serializeMap(expectedEntrypointsToDeps));
     });
+
```
usergenic: This test doesn't prove anything was broken.  I wrote it to demonstrate this case was not broken, to help zero in on the root cause.

CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+- When injecting html imports into documents for inlining, bundler now tries
```
justinfagnani: "When inlining HTML imports, preserve the ordering of imports between inlined and referenced documents"?

src/ast-utils.ts
```diff
@@ -114,6 +114,28 @@ export function parse(html: string, options: ParserOptions): ASTNode {
...
 
 /**
+ * Returns true if the nodes are given in order as they appear in the source
+ * code.
+ */
+export function inOrder(left: ASTNode, right: ASTNode): boolean {
+  const l = left.__location, r = right.__location;
+  return l && r && l['line'] && r['line'] &&
+      (l['line'] < r['line'] ||
+       (l['line'] === r['line'] && l['col'] < r['col']));
```
justinfagnani: nit: indentation alignment?

src/ast-utils.ts
```diff
@@ -114,6 +114,28 @@ export function parse(html: string, options: ParserOptions): ASTNode {
 }
 
 /**
+ * Returns true if the nodes are given in order as they appear in the source
+ * code.
+ */
+export function inOrder(left: ASTNode, right: ASTNode): boolean {
+  const l = left.__location, r = right.__location;
+  return l && r && l['line'] && r['line'] &&
```
justinfagnani: we don't have better typing available here?

CHANGELOG.md
usergenic: I guess the notion of `injecting` is outside what users might conceive of.  I'll reword.

A note: evaluation ordering of imports can not be 100% preserved in cases where merge strategies build shared bundles -- I could contrive examples where it wont be, in tests, but not in this PR (like document X having 5 links to files that have since been moved into a shared bundle Y, now document X just links to shared bundle Y once and the order of links in Y is unknown).  The aim is just to preserve the order where it has been expressed in the dependency graph.

src/ast-utils.ts
usergenic: Apparently not.  I could add typings for `LocationInfo` and `ElementLocationInfo` but I guess I'd rather do that after porting this to `dom5`.  I'll add a TODO.

src/ast-utils.ts
usergenic: This is what clang-format said.  I think its because of the containing `(`

CHANGELOG.md
usergenic: Updated CHANGELOG description to:
```
- In cases such as the shell merge strategy, imports which are not originally
  part of the shell are now inserted before other imports which are dependent
  on them.  Fixes https://github.com/Polymer/polymer-bundler/issues/517.
```

src/ast-utils.ts
```diff
@@ -114,6 +114,32 @@ export function parse(html: string, options: ParserOptions): ASTNode {
 }
 
 /**
+ * Returns true if the nodes are given in order as they appear in the source
+ * code.
+ * TODO(usergenic): Port this to `dom5` and do it with typings for location info
+ * instead of all of these string-based lookups.
+ */
+export function inOrder(left: ASTNode, right: ASTNode): boolean {
```
aomarks: maybe inSourceOrder?

src/bundler.ts
```diff
@@ -281,6 +254,17 @@ export class Bundler {
   }
 
   /**
+   * Append a <link rel="import" node to `node` with a value of `url` for
```
aomarks: missing `>` in the link tag

src/bundler.ts
```diff
@@ -324,6 +308,83 @@ export class Bundler {
...
+   * ensure that imports are injected prior to any eager imports of other
+   * bundles which are known to depend on them, to preserve expectations of
+   * evaluation order.
+   */
+  private _injectHtmlImportsForBundle(
+      document: Document,
+      ast: ASTNode,
+      bundle: AssignedBundle,
+      bundleManifest: BundleManifest) {
+    // Gather all the document's direct html imports.  We want the direct (not
```
aomarks: Nice commenting :)

src/ast-utils.ts
usergenic: 👍

src/bundler.ts
usergenic: 👍

PR #539, opened 2017-06-08
fix #531: URLs in inlined stylesheets are rewritten incorrectly when they refer out of the build root

### LINKED ISSUE #531: URLs in inlined stylesheets are rewritten incorrectly when they refer out of the build root
When polymer-bundler inlines styles with URLs (e.g. background images or fonts), the URLs referring to the resources outside of the build root are rewritten incorrectly.

**Example:**
With the following folder structure:
```
webapp/frontend/styles.css
webapp/frontend/index.html
webapp/images/icon.svg
```

Run the polymer-bundler in the webapp/frontend folder:
```
$ cd webapp/frontend
$ polymer-bundler index.html
```

See that in the output HTML the URL from the styles.css is incorrectly rewritten to `images/icon.svg` whereas the URL in the inline style block remained correct (`../images/icon.svg`)

index.html:
```html
<html lang="en">
<head>
  <meta charset="UTF-8">
  <link rel="stylesheet" href="styles.css">
  <title>Unbundled Resource</title>
</head>
<body>
<style type="text/css">
  .from-style-in-doc-body {
    background: url(../images/icon.svg);
  }
</style>
</body>
</html>
```

styles.css:
```css
.from-style-in-linked-css {
  background: url(../images/icon.svg);
}
```

polymer-bundler output:
```html
<!DOCTYPE html><html lang="en"><head>
  <meta charset="UTF-8">
  <style>.from-style-in-linked-css {
  background: url("images/icon.svg");
}
</style>
  <title>Unbundled Resource</title>
</head>
<body>
<style type="text/css">
  .from-style-in-doc-body {
    background: url(../images/icon.svg);
  }
</style>
</body></html>
```

**Rationale for using resources from outside of the build root:**
We build and ship several versions of the frontend code (es5 and es6) inside our deployment package. While the JS / HTML bundles are different in each build, the static resources (like fonts and images) are the same. In order to avoid duplicating that static content, we keep them out of the build root and reference with relative URLs from the files within the build root.

vlukashov:
This fix applies only when both the old and the new base URLs are relative (because otherwise referring outside of the build root is incorrect):
- add a test verifying the relative URL rewrites handles out-of-build-root URLs
- add a special handling for the case when both the new and the old base URLs are relative and run all path-related logic in relative paths then

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- Fixed issue with incorrect URL rewrites in inlined CSS (when URL points outside of the package root). https://github.com/Polymer/polymer-bundler/issues/531
 <!-- Add new, unreleased changes here. -->
 
 ## 2.0.3 - 2017-06-07
--- a/src/test/url-utils_test.ts
+++ b/src/test/url-utils_test.ts
@@ -100,6 +100,15 @@ suite('URL Utils', () => {
       testRewrite(
           'foo.html', 'bar.html', 'index.html', 'foo.html', 'neither has ^/');
     });
+
+    test('Rewrite URLs pointing outside of the build root', () => {
+      testRewrite(
+          '../../images/icon.svg',
+          'css/foo.css',
+          'foo.html',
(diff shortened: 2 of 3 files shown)


usergenic:
This patch looks good.  I'm going to merge into a branch and verify tests before merging into master.  Thanks, @vlukashov

PR #540, opened 2017-06-09
Added a --root|-r option to set the project root when calling polymer-bundler CLI.

usergenic:
- Now you can set the project root instead of it just always being current working directory.


usergenic:
cc @freshp86

justinfagnani:
@usergenic @aomarks How does this relate to the `root` setting/flag in the CLI here: https://github.com/Polymer/polymer-cli/blob/master/src/args.ts#L48

Issue #541, opened 2017-06-09
--strip-comments misses comments inside <template> (compared to Vulcanize)


freshp86:
Example file bug.html

```html
<!DOCTYPE html>
<html>
  <body>
    <!-- Top-level comment -->
    <style>
     /* comment inside <style>*/
    </style>
    <template>
      <!-- comment inside <template> -->
    </template>
    <div>
      <!-- comment inside <div> -->
    </div>
  </body>
</html>
```
Executing vulcanize
```
./node_modules/vulcanize/bin/vulcanize --strip-comments bug.html
```
outputs
```html
<!DOCTYPE html><html><head><meta charset="UTF-8"></head><body>
    
    <style>
     /* comment inside <style>*/
    </style>
    <template>
      
    </template>
    <div>
      
    </div>
  

</body></html>
```
Note that all comments except the one inside the `<style>` tag were removed. On the other hand using polymer-bundler as follows
```
./node_modules/polymer-bundler/lib/bin/polymer-bundler.js --strip-comments --in-html bug.html
```
outputs
```html
<!DOCTYPE html><html><body>
    <style>
     /* comment inside <style>*/
    </style>
    <template>
      <!-- comment inside <template> -->
    </template>
    <div>
      </div>
  

</body></html>
```
Note that the comment inside the `<template>` was not removed. Is that behavioral change intentional?

/cc @usergenic @danbeam

usergenic:
@freshp86 PR #542 fixes this.

PR #542, opened 2017-06-12
Update stripComments logic to include comments in template tags.

usergenic:
- Failing test in commit a697952 https://travis-ci.org/Polymer/polymer-bundler/jobs/242103581
- Fixed in commit 21be23e
- Fixes https://github.com/Polymer/polymer-bundler/issues/541
- [x] CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 - Added a `--redirect` option to `bin/polymer-bundler`.
 - Added a `--root` option to `bin/polymer-bundler`.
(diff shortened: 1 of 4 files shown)


stramel:
Does this preserve `<!--!` comments?

usergenic:
@stramel Currently there are two types of comments preserved:

1.) server-side include style `<!--# ... -->`
2.) comments with `@license` tag inside.

What is `<!--!` ?

usergenic:
@stramel since this PR doesn't change the definition of the type of comments we strip, I'm inclined to merge as-is.  If you want to preserve `<!--! ... --->` comments, we can probably add that, but lets add issue for it.

PR #544, opened 2017-06-12
Don't add the `original` property to a mapping unless one exists (#519)

### LINKED ISSUE #516: Sourcemaps don't work

usergenic:
- Fixed an error when generating source maps where `original.line` and/or `original.column` were not given.
 - Don't add the `original` property to a mapping unless one exists. This is needed for mappings to pass the `_validateMapping` check in https://github.com/mozilla/source-map/blob/master/lib/source-map-generator.js#L264
 - Fixes #516

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- Fixed error when generating source maps due to original line or column not being available.
 - Added a `--redirect` option to `bin/polymer-bundler`.
 - Added a `--root` option to `bin/polymer-bundler`.
 - The `stripComments` behavior now removes comments inside templates.
--- a/src/source-map.ts
+++ b/src/source-map.ts
@@ -83,13 +83,16 @@ function offsetSourceMap(
   consumer.eachMapping(mapping => {
     const newMapping: any = {
       source: mapping.source,
-      original: {line: mapping.originalLine, column: mapping.originalColumn},
       generated: {
         line: mapping.generatedLine + lineOffset,
         column: mapping.generatedColumn +
             (mapping.generatedLine === 1 ? firstLineCharOffset : 0)
       }
     };
+
+    if (typeof mapping.originalLine === 'number' && typeof mapping.originalColumn === 'number') {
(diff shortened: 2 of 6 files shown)


usergenic:
Thanks to @Lange for the fix.

usergenic:
Thanks, @aomarks

Issue #549, opened 2017-06-13
Extend `--manifest-out` option to include ALL inlined files, not just html imports


usergenic:
Right now, inlined scripts and css are not included in the `--manifest-out` content.  Lets include them.

usergenic:
cc @freshp86

Issue #550, opened 2017-06-13
The `--manifest-out` option should output a manifest even in a 1 bundle scenario.


usergenic:
The part where we output the bundle manifest:
https://github.com/Polymer/polymer-bundler/blob/master/src/bin/polymer-bundler.ts#L277

...is inside a check for `if (bundles.size > 1) {`

No!

usergenic:
cc @freshp86

usergenic:
Fixed in #551

PR #551, opened 2017-06-13
Make --manifest-out work with single bundle case and include js + css assets too.

usergenic:
- Fixes https://github.com/Polymer/polymer-bundler/issues/549 and https://github.com/Polymer/polymer-bundler/issues/550

### DIFF
--- a/src/test/polymer-bundler_test.ts
+++ b/src/test/polymer-bundler_test.ts
@@ -51,7 +51,7 @@ suite('polymer-bundler CLI', () => {
 
     test('writes out the bundle manifest to given path', async () => {
       const projectRoot = path.resolve(__dirname, '../../test/html');
-      const tempdir = fs.mkdtempSync(os.tmpdir());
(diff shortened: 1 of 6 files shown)


src/test/polymer-bundler_test.ts
```diff
@@ -46,6 +47,30 @@ suite('polymer-bundler CLI', () => {
     assert.include(stdout, 'hello from /absolute-paths/script.js');
   });
 
+  suite('--manifest-out', () => {
+
+    test('writes out the bundle manifest to given path', async () => {
+      const projectRoot = path.resolve(__dirname, '../../test/html');
+      const tempdir = tempy.directory();
```
aomarks: Can you use https://nodejs.org/api/fs.html#fs_fs_mkdtempsync_prefix_options instead?

usergenic: Cool!  Didn't know about that.  Converting.

Issue #530, opened 2017-06-04
Honor the `exclude` argument when inlining styles and scripts


vlukashov:
When using the 'excludes' option I want it to work with styles and scripts as well as with HTML imports. Currently it seems to work only with HTML imports.

**Example 1 - CSS inlining does not honor the 'exclude' option:**
(the following 2 commands yield the same result)
```bash
$ cd polymer-bundler/test/html
$ polymer-bundler --in-html inline-styles.html
```
```bash
$ cd polymer-bundler/test/html
$ polymer-bundler --exclude "imports/regular-style.css" --in-html inline-styles.html
```
The contents of `imports/regular-style.css` is inlined in both cases, while the expected behavior is that in the second case the link to `imports/regular-style.css` remains a link.

**Example 2 - scripts inlining does not honor the 'exclude' option:**
(the following 2 commands yield the same result)
```bash
$ cd polymer-bundler/test/html
$ polymer-bundler --in-html scripts.html
```
```bash
$ cd polymer-bundler/test/html
$ polymer-bundler --exclude "external/external.js" --in-html scripts.html
```
The contents of `external/external.js` is inlined in both cases, while the expected behavior is that in the second case the link to `external/external.js` remains a link.

vlukashov:
Would you accept a PR for this?

ankon:
This is quite important in relation to https://github.com/webcomponents/webcomponentsjs/issues/798 - essentially one cannot use webcomponents-loader.js when using polymer-bundler, because bundler nukes out the script, and that breaks the loader's functionality to load needed polyfills.

usergenic:
Fixed in #553 and being released in bundler 2.1.0

PR #553, opened 2017-06-14
Support exclusion of specific js and css files and containing folders.

### LINKED ISSUE #548: Excludes option should honor js and css assets in addition to html imports
The `excludes` option appears to only apply to html imports and folders containing those imports.  It needs to honor js and css assets as well.

usergenic:
- Fixes ~#548~ #530 


usergenic:
cc @freshp86

PR #558, opened 2017-06-15
Do not inline unless options are set in CLI

jscissr:
Fixed the `--inline-scripts` and `--inline-css` options, previously inlining happened regardless of whether the options were actually set.

Without the coercion to Boolean, the options were `undefined`, and
`undefined` is interpreted as `true` by the bundler because inlining is the
default.


usergenic:
@jscissr good catch.   We updated Bundler class to default to inlining scripts and css and  I incorrectly assumed `--inline-scripts=false` was accepted by the CLI arguments parser.  This is a great patch.  Accepted + merging.

Issue #562, opened 2017-06-19
stripUrlFileSearchAndHash() adds extra trailing slash


freshp86:
```js
stripUrlFileSearchAndHash('chrome://bookmarks/app.html'); // Returns chrome://bookmarks//
```

Note the double slash at the end.  This ends up in the final HTML bundle as

```html
<dom-module id="bookmarks-edit-dialog" assetpath="chrome://bookmarks//">
```

The code in question is at https://github.com/Polymer/polymer-bundler/blob/master/src/url-utils.ts#L46, and it seems that it could be easily fixed by using the [ensureTrailingSlash()](https://github.com/Polymer/polymer-bundler/blob/master/src/url-utils.ts#L31) helper a few lines above.

cc @usergenic

usergenic:
I am fairly certain this is a bug in node's abysmal url parsing library.  Prioritizing fix.

freshp86:
Perhaps the root cause is in node's URL parsing library. FWIW, I was able to fix this locally by changing url-utils.ts line 46 as follows
```js
u.pathname = path.posix.dirname(u.pathname + '_') + '/';
```

```js
u.pathname = ensureTrailingSlash(path.posix.dirname(u.pathname + '_'));
```

usergenic:
Fixed in #564

PR #564, opened 2017-06-20
Ensure trailing slash

usergenic:


src/test/url-utils_test.ts
```diff
@@ -28,6 +28,12 @@ suite('URL Utils', () => {
 
   suite('stripUrlFileSearchAndHash', () => {
 
+    test('Strips "man.html" basename off url', () => {
+      assert.equal(
+          urlUtils.stripUrlFileSearchAndHash('shark://alligator/man.html'),
+          'shark://alligator/');
```
usergenic: used to return `shark://alligator//` before the fix.

Issue #561, opened 2017-06-17
Unable to get document when using FSUrlLoader (polymer-analyzer)


VandeurenGlenn:
## code

```js
const analyzer = new Analyzer({
    urlLoader: new FSUrlLoader(join(process.cwd(), 'build/unbundled'))
});

const bundler = new polymerBundler.Bundler(config);

bundler.generateManifest('index.html').then(function (manifest) {
  console.log(manifest); // empty BundleManifest
  bundler.bundle(manifest).then(function (result) {
     console.log(result.documents.get('index.html').ast)
  });
  });
});
```

## result
![image](https://user-images.githubusercontent.com/8685280/27254658-0d4005c0-538d-11e7-8fd8-d65c2c6a9d7a.png)

usergenic:
`bundler.generateManifest(entrypoints: UrlString[])` <-- needs an array

Try:
`bundler.generateManifest(['index.html'])` ...

usergenic:
I'll file an issue to run some input checks on public API surface.  We rely a lot on TypeScript internally but want to guard against type errors in JS too.

VandeurenGlenn:
Actually I didn't quite explain good enough, the input is an array already and everything works, but when adding the analyzer It doesn't anymore and gives back above result

@usergenic

usergenic:
hmm.  can you output/confirm the result of `join(process.cwd(), 'build/unbundled')`

Also, I don't see the analyzer in the config (I assume that happened elsewhere?)

Suggest perhaps:
```javascript
const analyzer = new Analyzer({
    urlLoader: new FSUrlLoader(path.resolve('build/unbundled'))
});

const bundler = new polymerBundler.Bundler({
    analyzer: analyzer
});

bundler.generateManifest(['index.html']).then(function (manifest) {
  console.log(manifest); // empty BundleManifest
  bundler.bundle(manifest).then(function (result) {
     console.log(result.documents.get('index.html').ast)
  });
});
```

VandeurenGlenn:
The path is always as expected (D:\workspace\webup\build\unbundled).
I also tested multiple solutions:  join, resolve or just 'build/unbundled' all results are the same.
When setting a wrong path, an error is given that the file isn't found (as expected).

In the project its used in 2 different places
- [bundler](https://github.com/VandeurenGlenn/webup/blob/master/src/utils/bundler.js#L16)

[...]

PR #567, opened 2017-06-27
polymer-analyzer version mismatch stopgap.

usergenic:
- Better error messaging when a different version of polymer-analyzer is used in bundler's constructor.  A bundler-specific fix for what https://github.com/Polymer/polymer-analyzer/pull/686 tries to fix in polymer-analyzer itself.


usergenic:
Many users face essentially the same symptom as this https://github.com/Polymer/polymer-bundler/issues/561 which comes down to multiple versions of polymer-analyzer in their node_modules folder.  This is a stopgap fix for that.

Thanks, @aomarks

Issue #560, opened 2017-06-17
--out-dir is ignored if only one --in-html is provided


freshp86:
Minimal repro: Put the following contents in bug.html

```html
<div>Hello bug.html</div>
```

Then execute
```
node node_modules/polymer-bundler/lib/bin/polymer-bundler.js --in-html bug.html --out-dir foo/
```

The output is thrown in stdout, and no `foo/` folder is created. If a 2nd HTML file is passed with `--in-html` the output folder is created and stdout is empty as expected.

freshp86:
cc @usergenic

freshp86:
Related code seems to be at https://github.com/Polymer/polymer-bundler/blob/master/src/bin/polymer-bundler.ts#L282-L299.

freshp86:
Should this be marked as fixed?

usergenic:
Yep.  Closing.  Fixed.

Issue #559, opened 2017-06-17
--out-dir option for bin/polymer-bundler doesn't seem to handle absolute paths


usergenic:
This should be as easy as changing the line:
```typescript
const out = pathLib.join(process.cwd(), outDir, url);
```
to
```typescript
const out = pathLib.resolve(pathLib.join(outDir, url));
```

usergenic:
cc @freshp86

PR #571, opened 2017-06-28
Update CHANGELOG and package-lock.json for 2.1.1 release.

usergenic:
## 2.1.1 - 2017-06-28
 - Fixed the `--inline-scripts` and `--inline-css` options, previously inlining happened regardless of whether the options were actually set.  Important, if you use the `bin/polymer-bundler` CLI you must provide these options to inline scripts and css now.
 - Fixed issue where absolute paths were not handled properly. https://github.com/Polymer/polymer-bundler/issues/559
 - Fixed issue where out-dir was essentially ignored for a single bundle case. https://github.com/Polymer/polymer-bundler/issues/560
 - Fixed issue where, if a different version of polymer-analyzer was given to Bundler's constructor, it caused instanceof check failures and resulting in documents not being identified as documents; added more definitive error messaging for this scenario as stop-gap until we switch to a no instanceof version of polymer-analyzer interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 <!-- Add new, unreleased changes here. -->
 
-## 2.2.0 - 2017-06-28
+## 2.1.1 - 2017-06-28
 - Fixed the `--inline-scripts` and `--inline-css` options, previously inlining happened regardless of whether the options were actually set.  Important, if you use the `bin/polymer-bundler` CLI you must provide these options to inline scripts and css now.
 - Fixed issue producing assetpath in application root. https://github.com/Polymer/polymer-bundler/issues/562
 - Fixed issue where absolute paths were not handled properly. https://github.com/Polymer/polymer-bundler/issues/559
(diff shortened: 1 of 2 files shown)


usergenic:
NOTE wrt to the `--inline-scripts` and `--inline-css` arguments to the polymer-bundler CLI:  The default behavior of the Bundler class is to inline scripts and css.  The CLI flags don't support a negation like `--inline-scripts=false` and I didn't want to make changes to use a flag like `--no-inline-scripts` etc, so the CLI differs from the default behavior of the bundler class in this way but I don't expect this to be a significant schism; just wanted to mention it.

katejeffreys:
Thanks for adding me to the reviewers list! Does this 

> The default behavior of the Bundler class is to inline scripts and css[...] the CLI differs from the default behavior of the bundler class in this way[...]

need to be mentioned in the README? If so do you want me to add it?

usergenic:
Thanks for review, @rictic -- I think I can probably get https://github.com/Polymer/polymer-bundler/pull/572 merged today (it was much easier to fix this than I thought it would be) so I'll actually try to release bundler with that as 2.2.0 instead of this as 2.1.1.  Closing this PR.

usergenic:
@katejeffreys I'll go ahead and update the README in release PR.  Also, totes thanks for your reply!  As I mentioned to peter I'll send a new release PR after #572 is merged in later today.  🤞

PR #572, opened 2017-06-28
Don't import the shell from fragments/entrypoints.

### LINKED ISSUE #471: Bundles in Shell strategy should not link to the Shell.
Currently the Bundler treats a Shell as just-another-bundle which is mostly correct; the problem is that the Shell, being already loaded prior to other bundles is should not be linked to from within other bundles.

It happens whenever a bundle's shared import is bundled into the Shell and rewriting of the html import occurs:

Entrypoint/fragment before bundling:
```html
<link rel="import" href="some/shared/dependency.html">
<link rel="import" href="other/shared/dependency.html">
```

Entrypoint/fragment after bundling:
```html
<!-- after bundling -->
<link rel="import" href="shell.html">
```

The effect of an unnecessary import link to the Shell is generally negligible/idempotent/mostly-irrelevant from the browser's perspective, since the browser should notice the Shell is already loaded and it will skip the duplicate link.

However, in the case of generating the push-manifest in the CLI, it turned out to need a specific workaround to prevent all fragments from including preload declarations for the Shell itself.

Before fixing this explicitly as a function of a 'shell' being defined, it might be worth considering whether we can rely on the lazy-import declarations to build a directed graph which we could use to exclude/strip back-linking.  I.e. if bundle A is always assumed to be loaded before bundle B then any shared dependencies B references that are part of bundle A can be stripped out of B at bundle time instead of rewritten to reference bundle A.

### LINKED ISSUE #489: Shell's import gets included in all the bundles
Suppose we have an application having **4 entrypoints** (named Entrypoint 1, 2, 3, 4).
There dependencies are as flow

### Dependency tree

Entrypoint | Dependencies
------------ | -------------
Entrypoint 1 | Dep1.html, common.html 
Entrypoint 2 | Dep2.html, common.html
Entrypoint 3 | common.html
Entrypoint 4 | Dep1.html

Now, with **merge strategy set to 3** content of each file will be

### Output content

Entrypoint | element code | imports
------------ | ------------- | ---------------
Entrypoint 1 | entrypoint 1 | shell, shared_bundle_1
Entrypoint 2 | entrypoint 2, dep 2 | shell
Entrypoint 3 | entrypoint 3 | shell
Entrypoint 4 | entrypoint 4 | shared_bundle_1
shell | common | -
shared_bundle_1 | dep 1 | shell

Now, if user opens **Entrypoint 4** files from _shell_ will also be loaded even though they are not required by _Entrypoint 4_.

Ideally, none of the _shared_bundle_ should have import to shell as required _Entrypoint_ will always have it.

usergenic:
- Fixes #471 / #489 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- Fixes issue where bundles produced by shell strategy incorrectly included eager html imports of the shell. https://github.com/Polymer/polymer-bundler/issues/471
+- Fixed the `--inline-scripts` and `--inline-css` options, previously inlining happened regardless of whether the options were actually set.  Important, if you use the `bin/polymer-bundler` CLI you must provide these options to inline scripts and css now.
 - Fixed issue producing assetpath in application root. https://github.com/Polymer/polymer-bundler/issues/562
 - Fixed issue where absolute paths were not handled properly. https://github.com/Polymer/polymer-bundler/issues/559
 - Fixed issue where out-dir was essentially ignored for a single bundle case. https://github.com/Polymer/polymer-bundler/issues/560
 - Fixed issue where, if a different version of polymer-analyzer was given to Bundler's constructor, it caused instanceof check failures and resulting in documents not being identified as documents; added more definitive error messaging for this scenario as stop-gap until we switch to a no instanceof version of polymer-analyzer interface.
 <!-- Add new, unreleased changes here. -->
-- Fixed the `--inline-scripts` and `--inline-css` options, previously inlining happened regardless of whether the options were actually set
 
 ## 2.1.0 - 2017-06-14
 - Added a `--redirect` option to `bin/polymer-bundler`.
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -45,6 +45,9 @@ export class Bundle {
   // Set of all files included in the bundle.
   files: Set<UrlString>;
 
+  // Set of imports which should be removed when encountered.
+  stripImports = new Set<UrlString>();
+
   // These sets are updated as bundling occurs.
   inlinedHtmlImports = new Set<UrlString>();
   inlinedScripts = new Set<UrlString>();
@@ -250,7 +253,10 @@ export function generateShellMergeStrategy(
     generateEagerMergeStrategy(shell),
     generateMatchMergeStrategy(
         (b) => b.files.has(shell) ||
-            b.entrypoints.size >= minEntrypoints && !getBundleEntrypoint(b))
+            b.entrypoints.size >= minEntrypoints && !getBundleEntrypoint(b)),
+    // Don't link to the shell from other bundles.
+    (bundles) => bundles.map(
+        (b) => b.files.has(shell) ? b : b.stripImports.add(shell) && b),
   ]);
 }
 
--- a/src/bundle-manifest.ts
+++ b/src/bundle-manifest.ts
@@ -250,10 +250,19 @@ export function generateShellMergeStrategy(
     throw new Error(`Minimum entrypoints argument must be non-negative`);
   }
   return composeStrategies([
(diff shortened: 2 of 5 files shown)


usergenic:
cc @web-padawan

CHANGELOG.md
```diff
@@ -6,12 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- Fixes issue where bundles produced by shell strategy incorrectly included eager html imports of the shell. https://github.com/Polymer/polymer-bundler/issues/471
```
FredKSchott: lots here, we should really do a new release...

src/bundle-manifest.ts
```diff
@@ -250,7 +253,10 @@ export function generateShellMergeStrategy(
     generateEagerMergeStrategy(shell),
     generateMatchMergeStrategy(
         (b) => b.files.has(shell) ||
-            b.entrypoints.size >= minEntrypoints && !getBundleEntrypoint(b))
+            b.entrypoints.size >= minEntrypoints && !getBundleEntrypoint(b)),
```
FredKSchott: Lines 254-260 are realllllly hard to grok, can you simplify? Add curly brackets for the arrow function  `(b) => { }`? Name `b`? Define these two functions before passing them to `generateMatchMergeStrategy`?

CHANGELOG.md
```diff
@@ -6,12 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
...
+- Fixed the `--inline-scripts` and `--inline-css` options, previously inlining happened regardless of whether the options were actually set.  Important, if you use the `bin/polymer-bundler` CLI you must provide these options to inline scripts and css now.
```
usergenic: I just moved and extended this particular point of the changelog.  No code in this PR effected this bit about the `--inline-scripts/css` arguments to the bundler CLI.

src/bundler.ts
```diff
@@ -393,7 +393,7 @@ export class Bundler {
       ast: ASTNode,
       bundle: AssignedBundle,
       bundleManifest: BundleManifest) {
-    const visitedUrls = new Set<UrlString>();
+    const visitedUrls = new Set<UrlString>(bundle.bundle.stripImports);
```
FredKSchott: `visitedUrls` isn't really accurate anymore, since it's pre-populated with URLs that weren't actually visited (but you just want the side effect that "visited" currently conveys).

Not a huge deal, but could be more clear

justinfagnani: add a comment for why stripImports being visited gets us the behavior we want

src/bundle-manifest.ts
```diff
@@ -250,7 +253,10 @@ export function generateShellMergeStrategy(
...
+    // Don't link to the shell from other bundles.
+    (bundles) => bundles.map(
```
justinfagnani: make this a named function?

justinfagnani:
I think it's time to file a docs issue for the bundler to give them a heads up.

src/bundle-manifest.ts
usergenic: Done.  Added `generateNoBackLinkStrategy(urls: string[])`

src/bundle-manifest.ts
usergenic: Rejiggered and added comments to `generateShellMergeStrategy`.  Perhaps that will help?

CHANGELOG.md
usergenic: That's the intent.  As soon as this merges I am going to create a release PR for bundler.

src/bundler.ts
usergenic: I've decided to rename that variable and the parameter for the inlining function to `stripImports` since it is closely tied to that use case anyway.

usergenic: Actually I went and renamed the variable to `stripImports` as well as in the inlining function along with adjustments to comments.  "visited" is too abstract anyways.

usergenic:
@justinfagnani I'll add a docs issue and link to it from the release PR I'll generate as soon as this is merged in.

PR #573, opened 2017-06-28
Prepare release 2.2.0

usergenic:
Lots of goodies in 2.2.0!  (I updated the README per feedback in prior release PR https://github.com/Polymer/polymer-bundler/pull/571, @katejeffreys)

## 2.2.0 - 2017-06-28
- Fixed issue where bundles produced by shell strategy incorrectly included eager html imports of the shell. https://github.com/Polymer/polymer-bundler/issues/471
- Fixed the `--inline-scripts` and `--inline-css` options, previously inlining happened regardless of whether the options were actually set.  Important, if you use the `bin/polymer-bundler` CLI you must provide these options to inline scripts and css now.
- Fixed issue where absolute paths were not handled properly. https://github.com/Polymer/polymer-bundler/issues/559
- Fixed issue where out-dir was essentially ignored for a single bundle case. https://github.com/Polymer/polymer-bundler/issues/560
- Fixed issue where, if a different version of polymer-analyzer was given to Bundler's constructor, it caused instanceof check failures and resulting in documents not being identified as documents; added more definitive error messaging for this scenario as stop-gap until we switch to a no instanceof version of polymer-analyzer interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-- Fixes issue where bundles produced by shell strategy incorrectly included eager html imports of the shell. https://github.com/Polymer/polymer-bundler/issues/471
+<!-- Add new, unreleased changes here. -->
+
+## 2.2.0 - 2017-06-28
+- Fixed issue where bundles produced by shell strategy incorrectly included eager html imports of the shell. https://github.com/Polymer/polymer-bundler/issues/471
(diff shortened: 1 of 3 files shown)


usergenic:
cc @freshp86

usergenic:
cc @web-padawan

freshp86:
@usergenic: Will there be a need for a new release once https://github.com/Polymer/polymer-analyzer/issues/697 is addressed?

usergenic:
@freshp86 not 100% sure if package-lock.json will force the old locked version - After analyzer removes typescript an npm install should just get the latest polymer-analyzer version.  We may release a 2.2.1 anyway though

PR #577, opened 2017-07-07
Make sure inlined style imports are ordered correctly

### LINKED ISSUE #575: vulcanize 1.x - reverses inlined style order inside dom-module
The styles are prepended, but when there's more than one, that effectively reverses their sequence.
https://github.com/Polymer/polymer-bundler/blob/1.x/lib/vulcan.js#L404

Possibly as simple as reversing the order this collection is iterated over: https://github.com/Polymer/polymer-bundler/blob/1.x/lib/vulcan.js#L369

This effects users stuck on vulcanize 1.x only

azakus:
Previously, `<link rel="import" type="css">` styles were inlined by
prepending to the top of the `<dom-module>` template.

However, if there were multiple imports, the order would be reversed,
which can cause incorrect styling when overriding.

This commit makes sure the order remains correct for multiple style
imports.

Fixes #575

### DIFF
--- a/lib/vulcan.js
+++ b/lib/vulcan.js
@@ -82,6 +82,15 @@ function buildLoader(config) {
   return loader;
 }
 
+function nextSibling(node) {
+  var parentNode = node.parentNode;
+  if (!parentNode) {
+    return null;
+  }
+  var idx = parentNode.childNodes.indexOf(node);
+  return parentNode.childNodes[idx + 1] || null;
+}
+
 var Vulcan = function Vulcan(opts) {
     // implicitStrip should be true by default
   this.implicitStrip = opts.implicitStrip === undefined ? true : Boolean(opts.implicitStrip);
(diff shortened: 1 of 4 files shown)


lib/vulcan.js
```diff
@@ -400,8 +410,14 @@ Vulcan.prototype = {
...
               }
               dom5.remove(tag);
-              // put the style at the top of the dom-module's template
-              this.prepend(domTemplate.childNodes[0], style);
+              if (!lastPolymerExternalStyle) {
+                // put the style at the top of the dom-module's template
+                this.prepend(domTemplate.childNodes[0], style);
+              } else {
+                // put this style behind the last polymer external style
+                dom5.insertBefore(domTemplate.childNodes[0], nextSibling(lastPolymerExternalStyle), style);
```
kevinpschaaf: Seems like it would be easier to have the promises resolve with the style, and then in the `then` take the array of styles from `Promise.all` and append them in one go?

Or else just reverse the array of links to make inserting easier.

azakus: I'm really not inclined to go refactor this deprecated branch too much.

stramel:
I thought the `<link rel="import" type="css"` was deprecated?

azakus:
@stramel deprecated, but not removed for Polymer 1.x, nor for 2.x.

lib/vulcan.js
kevinpschaaf: Ok, LGTM then

PR #578, opened 2017-07-07
Port of logic fix from #577

azakus:
#577 fixes a logic bug for inlining `<link rel="import" type="css">` links where multiple will be inlined in reverse order.

This is a port of #577 to the master branch.


azakus:
@usergenic can you write the test for this? should be similar to #575

usergenic:
I'll add test; then we merge.

Issue #579, opened 2017-07-08
polymer-bundler breaks css image paths


jaichandra:
My application has below project structure:

```
my-project
* src
  * assets
    * bower_components
      * my-library
        * assets
          * texture.png
        * x-foo
          * x-foo.html
    * elements.html (list of imports pointing to elements in `my-library` folder)
```
I have a style with below css in `my-library` folder:
```
--paper-toolbar: {         
          border-image: url("../assets/texture.png") 1 0 0 0;
}
```
After I run `polymer-bundler` on elements.html, I see below url in the bundled file:
```
--paper-toolbar: {         
          border-image: url("bower_components/my-library/assets/texture.png") 1 0 0 0;
}
```

When I run this project locally, the final path that browser looks for is 
`http://localhost:4200/assets/bower_components/my-library/x-foo/bower_components/my-library/assets/texture.png`

I think the style urls should not be updated by `polymer-bundler`.

usergenic:
Are you using `polymer-bundler` on command-line or `polymer` (aka polymer-cli)?  Please provide any version info you can.  I'll try and reproduce your symptom.

jaichandra:
Here's the test repo for replicate above issue: https://github.com/jaichandra/polymer-bundler-test
I am on version 2.2.0.

usergenic:
@jaichandra  It appears that vulcanize and polymer-bundler have apparently been doing exactly the wrong thing here since... forever, with respect to rewriting the templated `<style>` tag css contents when inlining imports.

I am really surprised this hasn't been reported as a bug until now.  Also, we have four unit tests which are designed to assert it does precisely the current incorrect behavior.

👎 

Fixing this immediately.

brettpostin:
I _think_ I've just encountered the same issue using vulcanize 1.15.3.

Project structure:

app
-- elements
----- `my-elements.html` (bundle root to a bunch of imports including `icons.html`)
-- images
----- icons
-------- `icons.html` (a dom-module wrapped stylesheet)

After vulcanized the assetpath of the bundled dom-module is set to `../images/icons/` which is correct. However all of the `src()` references have also been rewritten from `foo.png` to `../images/icons/foo.png` making the browser look in `/images/images/icons/foo.png`.

I'm a little confused where `vulcanize` fits in with the new `polymer-bundler`? Are they completely separate tools now? Can a fix be applied to it or will i need to migrate to `polymer-bundler`?

usergenic:
[...]

PR #580, opened 2017-07-10
Fix invalid url-rewriting in inlined templated element style tags.

usergenic:
- Fixed [issue #579](https://github.com/Polymer/polymer-bundler/issues/579) where `url()` values inside `<style>` tags inside of `<dom-module>` tags of inlined html imports were rewritten without consideration of the module's `assetpath` property.
 - Fixed a related issue when stylesheet imports are inlined inside of a `<dom-module>` the url resolution now takes into consideration the `assetpath`.


usergenic:
cc @jaichandra

PR #589, opened 2017-07-14
Prepare vulcanize 1.2.0 release

usergenic:
# This is a VULCANIZE release, not POLYMER BUNDLER.

We are releasing a version of vulcanize to support users who for *all sorts of reasons* can't migrate their pipelines to polymer CLI or polymer-bundler yet, but need to bundle even polymer 2 projects.

## 1.2.0 - 2017-07-14
- Fix erroneously reverse-order inlining of style imports `<link rel="import" type="css">`.
- Added `--polymer2` flag that changes some of the rewriting behaviors to support the fact that Polymer 2.x honors the `assetpath` of `<dom-module>` when interpreting style urls:
  - Disables rewriting urls in inlined html imports containing `<style>` tags inside `<dom-module>` containers.
  - Include consideration of container `<dom-module>` `assetpath` property when rewriting urls inside inlined style imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+
+## 1.2.0 - 2017-07-14
 - Fix erroneously reverse-order inlining of style imports `<link rel="import" type="css">`.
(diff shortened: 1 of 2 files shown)


usergenic:
Note to docs peeps @katejeffreys @arthurevans there probably shouldn't be any external documentation updates needed for this maintenance release of vulcanize, since we really aren't looking to promote or support vulcanize for building polymer 2 projects in the long run.

PR #591, opened 2017-07-14
Updated CHANGELOG for a 3.0.0 polymer-bundler release.

usergenic:
- [x] CHANGELOG.md has been updated.
- [x] README.md has been updated.
- The reason this is a major release to 3.0.0 is due to the breaking change of defaulting to polymer 2.x behavior and mapping all old polymer 1.x vulcanize-style url rewriting preferences to the `rewriteUrlsInTemplates` option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+<!-- Add new, unreleased changes here. -->
+
+## 3.0.0 - 2017-07-14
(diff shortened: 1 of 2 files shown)


usergenic:
@katejeffreys and @arthurevans wrt docs: I am unaware of places where `polymer-bundler` library itself is specifically documented other than its own README.  Are you aware of any?

usergenic:
Plan is to release this 3.0.0 version with the breaking change and then update the 2.x version of polymer-cli to either:
- Default `rewriteUrlsInTemplates` to true, maintaining CLI support for polymer 1.x mirroring old vulcanize behavior.
- Default `rewriteUrlsInTemplates` to `true` if CLI can identify the version of Polymer in the project is 1.x.  That could be done statically by looking at `bower.json` (which may be unreliable if install-time resolution made user choose a `2.x` version) or by checking the version key of the `bower.json` within the installed polymer component folder.

Still working that out.

arthurevans:
Pretty sure we don't reference the `polymer-bundler` library anywhere in the docs—certainly not more than a passing mention.  We'll update docs when CLI 2.0 comes out. (That we'll need to coordinate on.)

Issue #600, opened 2017-08-22
Bundler is not inlining css when the <link> is inside the <template>


ergo:
Bundler is not inlining css when the <link> element is enclosed by the <template> tags.

Following repository contains reproductible test case:

https://github.com/ergo/polymer-build-inlining-bug

When building shared-styles file https://github.com/ergo/polymer-build-inlining-bug/blob/master/src/shared-styles.html#L15 if `<link rel="stylesheet" href="shared-styles.css">` is enclosed by `<template>` then build output does not contained inlined styles, removal of the tag results in correct behavior.

When using old `vulcanize` approach the assets were inlined correctly.

rictic:
Specifically this is a style module that references an external stylesheet, i.e. 

```html
<dom-module id="shared-styles">
  <template>
      <link rel="stylesheet" href="shared-styles.css">
  </template>
</dom-module>
```

It seems like we should inline the following into the toplevel document:

```html
<dom-module id="shared-styles">
  <template>
    <style>
      /* content of shared-styles.css here */
    </style>
  </template>
</dom-module>
```

ergo:
That is correct and this was what `vulcanize` did in the past.

usergenic:
Marking this as a bug because bundler is trying to have feature parity with vulcanize and this behavior is crucial to support bundling for stylesheets authored outside the template.  Prioritizing.

ergo:
Hello, are there any news on this?

usergenic:
@ergo soon!  I am working on some other pressing issues but I am hoping to get back on this by end of week or start of next week.  Thanks for your patience

PR #604, opened 2017-09-25
Inline external stylesheet links inside templates

usergenic:
- Fixed [issue #600](https://github.com/Polymer/polymer-bundler/issues/600) where `<link rel=stylesheet>` inside a `<template>` was not inlined.  If you want to exclude a specific stylesheet from inlining, you can add its path the the `excludes` option (or `--exclude` on command line).

### DIFF
--- a/test/html/external-in-template.html
+++ b/test/html/external-in-template.html
@@ -20,5 +20,12 @@
         <script src="external/external.js"></script>
       </template>
     </dom-module>
+
+    <dom-module id="my-other-element">
+      <template>
+        <link rel="stylesheet" href="external/external.css">
+        <script src="external/external.js"></script>
+      </template>
+    </dom-module>
   </body>
 </html>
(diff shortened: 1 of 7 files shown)


usergenic:
cc @ergo

test/html/external-in-template.html
```diff
@@ -0,0 +1,24 @@
...
+    subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>External Resources</title>
+  </head>
+  <body>
+    <dom-module id="my-element">
+      <template>
```
justinfagnani: Can you add a test that link to a .css file from two separate templates?

usergenic: I updated the test to demonstrate the same file is inlined in two places.

PR #620, opened 2017-12-13
Javascript Bundles!

usergenic:
- [ ] CHANGELOG.md has been updated

At present, the commits in this PR were all WIP/checkpoints of incremental work and there needs a lot of cleanup and comments here.  Here's a basic overview of what's going on:

1. Tried to split the HTML bundle production code from the JS bundle production code.  There's some overlap here because of bundling code for inline script tags, but in the overlap case I try to import those routines from the JS bundle producer.
2. I also tried to avoid directly invoking babylon from bundler, but that was more or less impossible without abusing the analyzer horribly to get what I needed.  So there are a few cases where I needlessly analyzeContext to get a document to get its ast that will probably just be replaced with babylon.parse calls soon.
3. rollup code is somewhat duplicated in the bundle-html-document.ts and bundle-js-module.ts files so I'll probably create a separate function for that bit of code after factoring it out a bit.

If you want to get a sense for what this work DOES or results in, check the test files:
- https://github.com/Polymer/polymer-bundler/compare/javascript-bundles?expand=1#diff-5ba69ed9efce5f8da998e359ad340e04
- https://github.com/Polymer/polymer-bundler/compare/javascript-bundles?expand=1#diff-44b6a04d0c13b481737b9452973665ac
- https://github.com/Polymer/polymer-bundler/compare/javascript-bundles?expand=1#diff-566baf25512ada040cc4cc65b28ff83c <-- hmm i should probably rename this file to bundle-html-document_test


usergenic:
Waiting on https://github.com/Polymer/polymer-analyzer/pull/797 to be released in order to support bundling of inline script modules where HTML document baseUrl differs from HTML document url due to base tag with href.

src/import-utils.ts
```diff
@@ -452,28 +456,85 @@ function rewriteElementAttrsBaseUrl(
...
+      CallExpression: {
+        enter(path: babelTraverse.NodePath) {
+          const callExp = path.node as babel.CallExpression;
+          const callee = callExp.callee as babel.Node;
+          const arg0 = callExp.arguments[0] as babel.StringLiteral;
+          if (callee && callee.type === 'Import' && arg0 &&
+              babel.isStringLiteral(arg0)) {
+            const originalSource = arg0.value;
+            const resolvedSource =
+                urlLib.resolve(oldBaseUrl, originalSource) as ResolvedUrl;
```
rictic: This won't work in the future. Should use a UrlResolver for this.

src/import-utils.ts
```diff
@@ -452,28 +456,85 @@ function rewriteElementAttrsBaseUrl(
...
+            }
+          }
+        }
+      },
+      ImportDeclaration: {
+        enter(path: babelTraverse.NodePath) {
+          const impDec = path.node as babel.ImportDeclaration;
+          const originalSource = impDec.source.value;
+          const resolvedSource =
+              urlLib.resolve(oldBaseUrl, originalSource) as ResolvedUrl;
```
rictic: ditto

src/test/bundle-html-document_test.ts
```diff
@@ -0,0 +1,324 @@
...
+    Promise<BundleResult> {
+      const bundlerOptions = Object.assign({}, options) as Options;
+      if (!bundlerOptions.analyzer) {
+        bundlerOptions.analyzer = new Analyzer({
+          urlLoader: new FSUrlLoader(path.resolve(root)),
+        });
+      }
+      const bundler = new Bundler(bundlerOptions);
+      return bundler.bundle(
+          await bundler.generateManifest(urls as ResolvedUrl[]));
```
rictic: Consider using a resolver here

src/test/bundle-js-module_test.ts
```diff
@@ -0,0 +1,270 @@
...
+    Promise<BundleResult> {
+      const bundlerOptions = Object.assign({}, options) as Options;
+      if (!bundlerOptions.analyzer) {
+        bundlerOptions.analyzer = new Analyzer({
+          urlLoader: new FSUrlLoader(path.resolve(root)),
+        });
+      }
+      const bundler = new Bundler(bundlerOptions);
+      return bundler.bundle(
+          await bundler.generateManifest(urls as ResolvedUrl[]));
```
rictic: ditto

src/url-utils.ts
```diff
@@ -81,34 +89,41 @@ function pathPosixRelative(from: string, to: string): string {
  * the `fromUri` to the `toUri`.  Function is URL aware, not path-aware, so
  * `/a/` is correctly treated as a folder path where `/a` is not.
  */
-export function relativeUrl(fromUri: UrlString, toUri: UrlString): UrlString {
+export function relativeUrl(
```
rictic: This is coming to UrlResolver.

src/url-utils.ts
```diff
@@ -17,39 +17,40 @@
...
 import constants from './constants';
 
 const sharedRelativeUrlProperties =
     ['protocol', 'slashes', 'auth', 'host', 'port', 'hostname'];
 
-/**
- * A string representing a URL.
- */
-export type UrlString = string;
+export type UrlString = ResolvedUrl | FileRelativeUrl | PackageRelativeUrl;
```
rictic: ?

usergenic: Some urls may be of any type in certain places.  I still want to indicate I expect them to be one of these in spirit.  For example, relativeUrl(source, target) needs source to be ResolvedUrl but target to be resolved or relative...  I guess I don't know enough about PackageRelativeUrl yet...  You caught me-- this UrlString change is a little uncooked.

src/url-utils.ts
usergenic: A possible yay!

package.json
```diff
@@ -38,7 +47,7 @@
     "rewire": "^2.5.2",
     "source-map-support": "^0.4.2",
     "tslint": "^3.15.1",
-    "typescript": "^2.2.0",
+    "typescript": "~2.5.0",
```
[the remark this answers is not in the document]
usergenic: One of the 2.6.x versions was causing errors on compile that I didn't want to wrestle with.  This will get adjusted before PR is done.

src/url-utils.ts
rictic: `PackageRelativeUrl` is pretty much now only used for urls passed in from APIs or the command line. They're just a way of expressing that this is a file that shouldn't be resolved relative to any other url. Hm, actually, we can express that in the type system can't we…

package.json
```diff
@@ -13,14 +13,23 @@
...
+    "babylon": "^6.18.0",
     "clone": "^2.1.0",
     "command-line-args": "^3.0.1",
     "command-line-usage": "^3.0.3",
     "dom5": "^2.2.0",
     "espree": "^3.5.2",
     "mkdirp": "^0.5.1",
     "parse5": "^2.2.2",
-    "polymer-analyzer": "^3.0.0-pre.2",
+    "polymer-analyzer": "^3.0.0-pre.4",
```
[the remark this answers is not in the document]
usergenic: Yeah just noticed PR is breaking on travis cuz of pre.5 changes too.  Locked at pre.4 for now.

src/url-utils.ts
rictic: https://github.com/Polymer/polymer-analyzer/pull/805 to clarify things a bit more using the type system

usergenic:
This will not be merged.  Has been superseded by time and an even more gigantic PR #637

PR #626, opened 2018-01-17
Update Bundler to use Polymer Analyzer v3.0.0-pre.10

usergenic:
- [BREAKING] Upgraded to use `polymer-analyzer` version `3.0.0-pre.10` which requires the `ResolvedUrl` type in nearly all places where URLs are exchanged and provides strict typing on URLs used globally.  Eliminated the local `UrlString` type.


package.json
```diff
@@ -20,16 +20,17 @@
...
+    "vscode-uri": "^1.0.1"
   },
   "devDependencies": {
     "@types/chai": "^3.4.30",
     "@types/clone": "^0.1.30",
     "@types/mocha": "^2.2.29",
     "@types/node": "^6.0.33",
     "@types/parse5": "^2.2.33",
-    "@types/source-map": "^0.5.0",
+    "@types/source-map": "=0.5.2",
```
usergenic: Latest release of `@types/source-map` is missing type definitions.  Pinning this to 0.5.2 until they publish a fix.

src/import-utils.ts
```diff
@@ -43,19 +41,21 @@ export async function inlineHtmlImport(
...
     docBundle: AssignedBundle,
     manifest: BundleManifest,
     enableSourcemaps: boolean,
     rewriteUrlsInTemplates?: boolean,
     excludes?: string[]) {
   const isLazy = dom5.getAttribute(linkTag, 'rel')!.match(/lazy-import/i);
   const rawImportUrl = dom5.getAttribute(linkTag, 'href')!;
-  const importUrl = urlLib.resolve(document.url, rawImportUrl);
-  if (!analyzer.canResolveUrl(importUrl)) {
+  const importUrl = isAbsolutePath(rawImportUrl) ?
```
justinfagnani: Looks like this happens 3 times. Make a function?

src/url-utils.ts
```diff
@@ -19,50 +19,50 @@ import * as path from 'path';
...
     ['protocol', 'slashes', 'auth', 'host', 'port', 'hostname'];
 
 /**
  * A string representing a URL.
  */
-export type UrlString = string;
-
-export function ensureTrailingSlash(href: UrlString): UrlString {
-  return href.endsWith('/') ? href : href + '/';
+export function ensureTrailingSlash(href: string) {
```
justinfagnani: Update comment?

src/bin/polymer-bundler.ts
```diff
@@ -172,8 +171,10 @@ const usage = [
 const options = commandLineArgs(optionDefinitions);
 const projectRoot =
     options.root ? pathLib.resolve(options.root) : pathLib.resolve('.');
+const projectRootUrl =
+    ensureTrailingSlash(Uri.file(projectRoot).toString()) as ResolvedUrl;
```
justinfagnani: consider moving to a named function

usergenic: Done.  Created `getFileUrl` to dry it up.  Also created an analog to `path.resolve()` called `resolvePath` which preserves trailing slashes.  Converted all the places I call `ensureTrailingSlash`, `path.resolve()` and `Uri.file` to use the functions where possible.

src/bin/polymer-bundler.ts
```diff
@@ -224,59 +228,94 @@ if (options.redirect) {
             return {prefix, path};
           })
           .filter((r: redirection) => r.prefix && r.path);
-  const resolvers: UrlResolver[] =
-      redirections.map((r: redirection) => new RedirectionResolver(r.prefix));
+  const resolvers: UrlResolver[] = redirections.map((r: redirection) => {
+    let resolvedPath = Uri.file(path.resolve(r.path)).toString();
+    // Ensure trailing slash if input path had trailing slash
```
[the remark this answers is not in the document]
usergenic: This is now handled automatically by `resolvePath()`/`getFileUrl()` functions.

src/bin/polymer-bundler.ts
```diff
@@ -224,59 +228,94 @@ if (options.redirect) {
...
+    return new RedirectResolver(
+        packageUrlResolver.resolve(r.prefix as PackageRelativeUrl)!,
+        r.prefix,
+        resolvedPath);
+  });
+
+  // Wrap the getPackageRelativeUrl function for each redirection.  Because we
+  // are wrapping the function, we iterate in reverse order to ensure the
+  // redirect transformations are processed in the same order as they're given.
+  redirections.reverse().forEach((r: redirection) => {
```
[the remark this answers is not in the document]
usergenic: While I am delighted at this new method on Array I didn't know about, I'm not sure I can really apply it as I'm not actually reducing of redirection details, I'm composing a function.

However, per our discussion in person, this could be better if not generated as an n-sized onion of function calls but rather a single function with an internal iterator over the redirects etc.

For the time being I will leave this as-is and its going to be completely removed once the analyzer itself can provide a reliable PackageRelativeUrl for multiple redirects after I update `UrlResolver#relative()` to support undefined return values for fall-thru support in `MultiUrlResolver` context.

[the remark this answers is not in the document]
usergenic: Answered above in https://github.com/Polymer/polymer-bundler/pull/626#discussion_r163015284

src/import-utils.ts
usergenic: I delegate to `analyzer.urlResolver.resolve(document.url, rawImportUrl)` now which does this same job.

src/url-utils.ts
usergenic: Right.  Done.  Also updated type definition so that it can return the same type of `href` that comes in:

```ts
/**
 * Given a string representing a URL or path of some form, append a `/`
 * character if it doesn't already end with one.
 */
export function ensureTrailingSlash<T>(href: T): T {
  const hrefString = href as any as string;
  return hrefString.endsWith('/') ? href : (href + '/') as any as T;
}
```

src/bin/polymer-bundler.ts
```diff
@@ -224,59 +228,94 @@ if (options.redirect) {
...
-          new PrefixedUrlLoader(r.prefix, new FSUrlLoader(r.path)));
-  if (loaders.length > 0) {
+      (r: redirection) => new FSUrlLoader(path.resolve(r.path)));
+  if (redirections.length > 0) {
     options.analyzer = new Analyzer({
-      urlResolver:
-          new MultiUrlResolver([...resolvers, new PackageUrlResolver()]),
-      urlLoader:
-          new MultiUrlLoader([...loaders, new FSUrlLoader(projectRoot)])
+      urlResolver: new MultiUrlResolver([...resolvers, packageUrlResolver]),
```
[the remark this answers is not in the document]
usergenic: It doesn't technically get overwritten because it has no way to be set prior to here.  The options which are initialized in polymer-bundler.ts (which is the CLI) can not include an externally specified analyzer.

src/bin/polymer-bundler.ts
```diff
@@ -224,59 +228,94 @@ if (options.redirect) {
...
-    json[url] = [...new Set([
-      // `files` and `inlinedHtmlImports` will be partially duplicative, but use
-      // of both ensures the basis document for a file is included since there
-      // is no other specific property that currently expresses it.
-      ...bundle.files,
-      ...bundle.inlinedHtmlImports,
-      ...bundle.inlinedScripts,
-      ...bundle.inlinedStyles
-    ])];
+  // Produces a bundle manifest object where all the urls represented are
```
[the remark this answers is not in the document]
usergenic: I went ahead and did that everywhere in the repo #65ac55b

src/bin/polymer-bundler.ts
```diff
@@ -224,59 +228,94 @@ if (options.redirect) {
...
+  // package relative.
+  function bundleManifestToJson(manifest: BundleManifest): JsonManifest {
```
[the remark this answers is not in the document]
usergenic: It can except we need a way to get package relative URLs, so once I can get those from the Analyzer's UrlResolver, the resolver can be passed in as an argument and then I can move it to the BundleManifest.

usergenic:
@justinfagnani I've addressed existing PR feedback here.

package.json
usergenic: Verified this is still an issue.

PR #634, opened 2018-02-07
Refactor: Move all HTML-specific bundling code into new HtmlBundler class

usergenic:
## From the CHANGELOG:

## Details

This refactoring is a prelude to the introduction of upcoming class which will perform JavaScript module bundling.  It was necessary to create a new home just for the HTML bundling code since it will become unwieldy to mix concerns and the `import-utils` module was already ambiguously named and method signatures were getting too long (needing sometimes 6 parameters.)  It is much more concise to have the majority of these functions defined as methods on an HtmlBundler instance that can reach into bundler properties and get the objects and values they need.

There are *no behavioral changes* in this refactoring; this is just about renaming and moving stuff around.


src/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+/**
+ * Simple utility function used to find an item in a set with a predicate
+ * function.  Analagous to Array.find(), without requiring converting the set
+ * an Array.
+ */
+export function findInSet<T>(set: Set<T>, predicate: (item: T) => boolean): T|
```
aomarks: Could be even more generic with `Iterable<T>` I think?

src/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+    undefined {
+  for (const item of set) {
+    if (predicate(item)) {
+      return item;
+    }
+  }
+  return;
```
aomarks: don't need this return I guess?

src/test/html-bundler_test.ts
```diff
@@ -193,8 +209,9 @@ suite('import-utils', () => {
         </dom-module>
         <script>Polymer({is: "my-element"})</script>`;
 
-      const ast = astUtils.parse(htmlBase);
-      importUtils.rewriteAstToEmulateBaseTag(analyzer, ast, 'the/doc/url');
+      const ast = parse(htmlBase);
+      htmlBundler['_rewriteAstToEmulateBaseTag'](
```
aomarks: Why does this use `['_rewriteAstToEmulateBaseTag']` style now?

usergenic: oh because it's a private method.

usergenic: the original tests were testing behavior of exported functions in the import-utils module but now they're all methods on the HtmlBundler and only a few methods need to be public on that thing because of the way it will be used.  To preserve the tests without testing for new behavior, I just call into the private methods directly.

src/utils.ts
usergenic: Neat!  Doing that and renamed to `find`

src/utils.ts
usergenic: Removed.  Original code was going for being explicit but you're right it is not functional/meaningful at runtime.

PR #635, opened 2018-02-08
Added .type to Bundle class in preparation of JavaScript module type bundles.

usergenic:
- [BREAKING] `Bundle` class now has a required `type` property and as first argument to its constructor, with current possible values of `html-fragment` and `es6-module`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [BREAKING] Upgraded to use `polymer-analyzer` version `3.0.0-pre.10` which requires the `ResolvedUrl` type in nearly all places where URLs are exchanged and provides strict typing on URLs used globally.  Eliminated the local `UrlString` type.
 - [BREAKING] Removed `relativeUrl` function from `url-utils`, shifting responsibility to `Bundler#analyzer.urlResolver.relative()`.
 - [BREAKING] Removed `import-utils.ts` and migrated all HTML-specific bundling code out of it and out of `bundler.ts` into a new class `HtmlBundler`.  Renamed `ast-utils.ts` to `parse5-utils.ts` since "ast" is ambiguous when we have JavaScript ASTs coming in.
-- [BREAKING] `Bundle` class now has a required `type` property and as first argument to its constructor, with current possible values of `html-fragment` and `es6-module`.
+- [BREAKING] `Bundle` class now has a required `type` property as first argument to its constructor, with current possible values of `html-fragment` and `es6-module`.
 - The `BundleResult` returned by `Bundler#bundle()` includes serialized contents instead of just the root AST node of bundled document.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [BREAKING] Upgraded to use `polymer-analyzer` version `3.0.0-pre.10` which requires the `ResolvedUrl` type in nearly all places where URLs are exchanged and provides strict typing on URLs used globally.  Eliminated the local `UrlString` type.
 - [BREAKING] Removed `relativeUrl` function from `url-utils`, shifting responsibility to `Bundler#analyzer.urlResolver.relative()`.
 - [BREAKING] Removed `import-utils.ts` and migrated all HTML-specific bundling code out of it and out of `bundler.ts` into a new class `HtmlBundler`.  Renamed `ast-utils.ts` to `parse5-utils.ts` since "ast" is ambiguous when we have JavaScript ASTs coming in.
+- [BREAKING] `Bundle` class now has a required `type` property and as first argument to its constructor, with current possible values of `html-fragment` and `es6-module`.
```
aomarks: s/and as/as ?

src/bundle-manifest.ts
```diff
@@ -55,9 +55,13 @@ export class Bundle {
...
   inlinedStyles = new Set<ResolvedUrl>();
 
-  constructor(entrypoints?: Set<ResolvedUrl>, files?: Set<ResolvedUrl>) {
-    this.entrypoints = entrypoints || new Set<ResolvedUrl>();
-    this.files = files || new Set<ResolvedUrl>();
+  constructor(
+      // Filetype discriminator for Bundles.
+      public type: BundleType,
+      // Set of all dependant entrypoint URLs of this bundle.
+      public entrypoints: Set<ResolvedUrl> = new Set(),
```
aomarks: `= new Set<ResolvedUrl>` should work the same?

src/utils.ts
```diff
@@ -25,3 +25,15 @@ export function find<T>(items: Iterable<T>, predicate: (item: T) => boolean): T|
     }
   }
 }
+
+/**
+ * Returns a set of unique/distinct values returned by calling the given
+ * function on each item.
+ */
+export function uniq<T, R>(items: Iterable<T>, map: (item: T) => R): Set<R> {
```
aomarks: Is this equivalent to doing `new Set([...items].map((item)=>whatever(item)))`?

CHANGELOG.md
usergenic: Right! Done.

src/bundle-manifest.ts
usergenic: Right! Done.

src/utils.ts
usergenic: Yes... but this is a function I made!

PR #636, opened 2018-02-15
Extending DepsIndex and Bundle generation to support ES6 modules

usergenic:
- [BREAKING] The `DepsIndex` type returned from `buildDepsIndex` has changed to a single map of entrypoints to dependencies.
- Added support for `<script type="module">` and bundling of ES6 modules by following `import` statements.  Dynamic `import()` statements are treated similarly to the `<link rel="lazy-import">` tag in that the import target implicitly defines a new entrypoint/bundle.
- Shared bundles containing only one file are named after the original file instead of having naming delegated to shared bundle URL mapper.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [BREAKING] Removed `relativeUrl` function from `url-utils`, shifting responsibility to `Bundler#analyzer.urlResolver.relative()`.
 - [BREAKING] Removed `import-utils.ts` and migrated all HTML-specific bundling code out of it and out of `bundler.ts` into a new class `HtmlBundler`.  Renamed `ast-utils.ts` to `parse5-utils.ts` since "ast" is ambiguous when we have JavaScript ASTs coming in.
 - [BREAKING] `Bundle` class now has a required `type` property as first argument to its constructor, with current possible values of `html-fragment` and `es6-module`.
+- [BREAKING] The `DepsIndex` type returned from `buildDepsIndex` has changed to a single map of entrypoints to dependencies.
 - Added support for `<script type="module">` and bundling of ES6 modules by following `import` statements.  Dynamic `import()` statements are treated similarly to the `<link rel="lazy-import">` tag in that the import target implicitly defines a new entrypoint/bundle.
 - Shared bundles containing only one file are named after the original file instead of having naming delegated to shared bundle URL mapper.
 - The `BundleResult` returned by `Bundler#bundle()` includes serialized contents instead of just the root AST node of bundled document.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [BREAKING] Removed `relativeUrl` function from `url-utils`, shifting responsibility to `Bundler#analyzer.urlResolver.relative()`.
 - [BREAKING] Removed `import-utils.ts` and migrated all HTML-specific bundling code out of it and out of `bundler.ts` into a new class `HtmlBundler`.  Renamed `ast-utils.ts` to `parse5-utils.ts` since "ast" is ambiguous when we have JavaScript ASTs coming in.
 - [BREAKING] `Bundle` class now has a required `type` property as first argument to its constructor, with current possible values of `html-fragment` and `es6-module`.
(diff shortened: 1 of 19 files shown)


src/deps-index.ts
```diff
@@ -81,28 +29,170 @@ function _getHtmlDependencies(
...
+  function _getHtmlExternalModuleDependencies(document: Document) {
+    let externalModuleCount = 0;
+    const htmlScripts =
+        [...document.getFeatures({kind: 'html-script', ...getFeaturesOptions})]
+            .filter(
+                (i) => (i.document.parsedDocument as JavaScriptDocument)
+                           .parsedAsSourceType === 'module');
+    for (const htmlScript of htmlScripts) {
+      moduleScriptImports.set(
+          `external-module:${++externalModuleCount}`, htmlScript.document);
```
FredKSchott: how hard would it be to include information about the original `htmlScript.url` in this map key? This feels like useful info that shouldn't get lost, at the very least for debugging.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [BREAKING] Removed `relativeUrl` function from `url-utils`, shifting responsibility to `Bundler#analyzer.urlResolver.relative()`.
 - [BREAKING] Removed `import-utils.ts` and migrated all HTML-specific bundling code out of it and out of `bundler.ts` into a new class `HtmlBundler`.  Renamed `ast-utils.ts` to `parse5-utils.ts` since "ast" is ambiguous when we have JavaScript ASTs coming in.
 - [BREAKING] `Bundle` class now has a required `type` property as first argument to its constructor, with current possible values of `html-fragment` and `es6-module`.
+- [BREAKING] The `DepsIndex` type returned from `buildDepsIndex` has changed to a single map of entrypoints to dependencies.
+- Added support for `<script type="module">` and bundling of ES6 modules by following `import` statements.  Dynamic `import()` statements are treated similarly to the `<link rel="lazy-import">` tag in that the import target implicitly defines a new entrypoint/bundle.
+- Shared bundles containing only one file are named after the original file instead of having naming delegated to shared bundle URL mapper.
```
FredKSchott: Can you explain the reasoning behind this? This sounds like something that would may people up ("why was this file skipped by the bundler?")

src/bundle-manifest.ts
```diff
@@ -364,19 +421,26 @@ export function mergeBundles(bundles: Bundle[]): Bundle {
...
  */
 export function mergeMatchingBundles(
     bundles: Bundle[], predicate: (bundle: Bundle) => boolean): Bundle[] {
   const newBundles = Array.from(bundles);
   const bundlesToMerge = newBundles.filter(predicate);
-  if (bundlesToMerge.length > 1) {
-    for (const bundle of bundlesToMerge) {
-      newBundles.splice(newBundles.indexOf(bundle), 1);
+
+  const types = [...new Set(bundlesToMerge.map((b) => b.type))].sort();
```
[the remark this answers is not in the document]
usergenic: Done!

src/deps-index.ts
usergenic: Added!

src/bundle-manifest.ts
```diff
@@ -192,9 +194,58 @@ export function generateBundles(depsIndex: TransitiveDependenciesMap):
...
   }
+
   return bundles;
 }
 
+/**
+ * Instances of `<script type="module">` generate synthetic entrypoints in the
+ * depsIndex and are treated as entrypoints during the initial phase of
+ * `generateBundles`.  Before handing bundles to the strategy function, if
+ * script inlining is enabled, any bundle which provides dependencies to a
```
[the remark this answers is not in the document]
usergenic: Updated!

CHANGELOG.md
usergenic: Hmm.  The premise here was that it didn't make sense to essentially just rename a single file to 'shared_bundle_x.js' just because that one file was imported from two locations.  I'm realizing in thinking more about it that the single file case is actually more specific version of a broader class of 'single top-level direct dependency' expressed in the bundle, meaning that maybe 'a.js' is depended on by multiple entrypoints but 'a.js' imports 'b.js' and 'b.js' has no other direct dependents, meaning that there really could be a bundle called 'a.js' which rolls up 'b.js' into it, and that case is not handled here.

You've given me something more to think about.  Thanks!

usergenic: I removed this special-case logic and the CHANGELOG item for it.  It was one of those things that made sense in the moment but needs to be thought about and handled as a separate concern.

usergenic:
Closing this PR in favor of #637 mega PR

Issue #645, opened 2018-03-23
JavaScript bundling does not recognize default export if part of an ExportNamedDeclaration


usergenic:
Bundler understands `export default expression` form but not `export {x as default}` represents a default export is present.

usergenic:
This is fixed in https://github.com/Polymer/polymer-bundler/pull/647

Issue #641, opened 2018-03-12
Exports in "export * from" statements don't populate export list of bundled module


usergenic:
Bundler needs to be smarter about `export * from 'x'` when constructing its exported namespace values; right now it doesn't import 'x' and gather its exports.

usergenic:
This is fixed in https://github.com/Polymer/polymer-bundler/pull/647

PR #647, opened 2018-03-27
Use analyzers export features

usergenic:
Bundler had its own `getModuleExportNames` function which scanned for export identifiers.  Analyzer provides these now, so I changed code to delegate to Analyzer to obtain them.

This seemed to address two related issues:

 1. Identifying a default export in a form other than `export default x` such as `export {x as default}` https://github.com/Polymer/polymer-bundler/issues/645 
 2. Exports in "export * from" statements don't populate export list of bundled module https://github.com/Polymer/polymer-bundler/issues/641

So that's nice.

### DIFF
--- a/src/es6-module-bundler.ts
+++ b/src/es6-module-bundler.ts
@@ -19,7 +19,7 @@ import {getAnalysisDocument} from './analyzer-utils';
 import {AssignedBundle, BundleManifest} from './bundle-manifest';
 import {Bundler} from './bundler';
 import {BundledDocument} from './document-collection';
-import {getModuleExportNames, getOrSetBundleModuleExportName, hasDefaultModuleExport} from './es6-module-utils';
+import {getModuleExportNames, getOrSetBundleModuleExportName} from './es6-module-utils';
 import {Es6Rewriter} from './es6-rewriter';
 import {ensureLeadingDot, stripUrlFileSearchAndHash} from './url-utils';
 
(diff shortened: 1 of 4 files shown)


usergenic:
I added a test because there was a weird problem that I had to fix in latest analyzer around module export names including `default`.  That's fixed.  No logic changes though.

src/es6-module-bundler.ts
```diff
@@ -62,50 +62,37 @@ export async function bundle(
  *     export {$moduleA, $moduleB, $moduleBDefault};
  */
 async function prepareBundleModule(
```
usergenic: It looks like this entire block is actually indentation/formatting change due to clang-format or something IDK.

usergenic:
@justinfagnani best kind of PR is more lines removed than added :)  thx for review

src/test/es6-module-bundling-scenarios_test.ts
```diff
@@ -24,6 +24,45 @@ import {heredoc, inMemoryAnalyzer} from './test-utils';
...
+        B: B,
+        C: C
+      };
+      const A = 'a';
+      var a = {
+        A: A,
+        B: B,
+        C: C
+      };
+      export { a as $a, b as $b, c as $c, C, B, A, C as C$1, B as B$1, C as C$2, C as $cDefault };`);
```
rictic: If this is a representation of `a.js` then shouldn't we _not_ be exporting `$cDefault`?

usergenic: This is a simplifying bundler behavior that all modules that are part of a bundle export their exports from the bundle under a namespace for their module.

It's because we don't currently track which modules are depended on outside the bundle directly.  This is an optimization we may choose to make.  Filed: https://github.com/Polymer/polymer-bundler/issues/650

PR #651, opened 2018-04-03
Update external module script tags to reference bundles.

aomarks:


aomarks:
> LGTM.
> 
> It'd be nice to have test coverage of more scenarios, but let's make a list of tests to add as we do manual testing a discuss the variations to cover.

Yeah. There's also some variations in the CLI tests, but I think most of that coverage should indeed be here.

